PR #414, opened 2024-02-20
Fix/net8 build

dpavicic:


src/ModularPipelines/ModularPipelines.csproj
```diff
@@ -44,10 +44,5 @@
       <PackageReference Include="YamlDotNet" Version="15.1.0" />
     </ItemGroup>
 
-    <ItemGroup>
```
thomhurst, line 47: Hmm I don't want to remove this line. This makes dotnet pack aware that this is a dependency, so that when you install the base package, the analyzer is pulled in as well.

What issue are you getting by leaving this in?

If it is giving an issue locally, what we can do is add a condition to the item group so that this is brought in if we're in the context of the build pipeline (which is where the nuget packages are built and published.)

thomhurst: If so, the condition on the item group would be this:

Condition="'$(GITHUB_ACTIONS)' == 'true'"

This should mean it's brought in when building (and so makes it a dependency that will be pulled in) - but won't be pulled in locally, as that env variable will be null

dpavicic: Sorry, yes. These are just changes to be able to build locally. 

The actual change for .NET8 to work (and it minor) is here:
https://github.com/thomhurst/ModularPipelines/pull/414/commits/55a702bc48dac408b1cf02f25de8877dbb247f3c

src/ModularPipelines/Exceptions/PipelineException.cs
```diff
@@ -16,6 +16,9 @@ public PipelineException(string? message, Exception? innerException) : base(mess
     {
     }
 
+#if NET8_0_OR_GREATER
+    [Obsolete(DiagnosticId = "SYSLIB0051")] // add this attribute to the serialization ctor
+#endif
     protected PipelineException(SerializationInfo info, StreamingContext context) : base(info, context)
```
thomhurst, line 22: Actually, I don't think this constructor is ever used. All exceptions that inherit it just use a basic message or inner exception constructor.

Can you check this please? And if so, just remove this constructor completely. I think that's cleaner than keeping it in if nothing is using it at current. And that'll prevent people from using it in future if it's a now obsolete piece of code

dpavicic: Yep, works when removed.

src/ModularPipelines/ModularPipelines.csproj
thomhurst: What's the error you're getting out of curiosity?

PR #426, opened 2024-02-24
Make some internal classes public to allow for more customization through dependency injection and encaspulation.

ScionOfDesign:
I recently discovered ModularPipelines and have found it to be incredibly useful, especially its implementations of various command line programs. However, these implementations are coupled with the rest of the ModularPipelines code. This normally would not be a problem, but having the default implementations of the injected interfaces such as `Command` be internal is highly inconvenient for extensibility.

Summary of changes:
1. Make `ICommandLogger`, `ILogoPrinter` public.
    - Justification: Allow for customization of Command Execution and Logo Printing via Dependency Injection.
    - In my specific use case, I would like to customize the logo and execute commands through a class that will execute commands either locally or remotely through SSH.net.
3. Made `Command`, `CommandLogger`, and `LogoPrinter` public and sealed. Added a `Default` option to the enum `CommandLogging`.
    - Justification: Allow aforementioned customization to fall back to default behavior via encapsulation.
    - In my specific use case, I don't want to have to re-implement `Command` if I want to execute commands locally after injecting my own `ICommand`.
4. Added a `PrintLogo` option to `PipelineOptions`, allowing users to disable logo printing.
    - Justification: Currently, the only way to disable the logo is to set `AnsiConsole.Console.Profile.Capabilities.Interactive = false;`
    - This is inconvenient and an option for it should be present in the `PipelineOptions`.
5. Added a `Default` option to the `CommandLogging` enum.
    - Justification: Creating a single source of truth for the default `CommandLogging` options.


ScionOfDesign:
There might need to be some changes to `CommandResult` as well, it seems. So that the custom `ICommand` I wish to implement can return a valid CommandResult.

thomhurst:
Thanks @ScionOfDesign ! Changes are fine by me, I didn't even consider people wanting to extend/replace things like the logo printing.

Can you add some release notes? The build requires them to be added, so they're available when generating a new release.

Just a line or two in here regarding your changes: 
> src/ModularPipelines.Build/ReleaseNotes.md

ScionOfDesign:
I see you made some changes to `CommandResult` to add a public constructor. This looks good now!

thomhurst:
Thanks for the contribution @ScionOfDesign ! New version should be out. Let me know if it's okay.

PR #460, opened 2024-03-27
🐛 Don't Increment Module Progress if it will exceed 100

### LINKED ISSUE #459: 🐛 Interactive Console Progress - Module appears to finish before actually finished

MattParkerDev:
Closes #459

### DIFF
--- a/src/ModularPipelines/Helpers/ProgressPrinter.cs
+++ b/src/ModularPipelines/Helpers/ProgressPrinter.cs
@@ -168,9 +168,9 @@ internal class ProgressPrinter : IProgressPrinter
                 progressTask.StartTask();
                 var estimatedDuration = moduleToProcess.EstimatedDuration * 1.1; // Give 10% headroom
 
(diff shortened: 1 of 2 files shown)


MattParkerDev:
I'm happy to change 95 to 99 or whatever you'd prefer :)

thomhurst:
Nice one! Can you add a line to the ReleaseNotes.md in sec/ModularPipelines.Build?

It'll be useful when generating a new release.

src/ModularPipelines/Helpers/ProgressPrinter.cs
```diff
@@ -173,7 +173,7 @@ private static string GetModuleExtraInformation(ModuleBase module)
                 var ticksPerSecond = 100 / totalEstimatedSeconds;
```
thomhurst, line 173: Could you also turn this into 100.0? The fact it's an int is probably causing rounding issues which might be the reason for this.

MattParkerDev: done!

thomhurst:
Thanks again @MattParkerDev ! New version is out :smile:

PR #1693, opened 2025-12-30
feat: Enhance Folder.Clean() and CopyTo() methods with additional options

thomhurst:
## Summary
- Added `Clean(bool removeReadOnlyAttribute)` overload to handle read-only files
- Added `CopyTo(string targetPath, bool preserveTimestamps)` overload for timestamp control
- Both methods default to backward-compatible behavior:
  - `Clean()` defaults to `removeReadOnlyAttribute: false` (preserves original throwing behavior)
  - `CopyTo()` defaults to `preserveTimestamps: false` (preserves original OS-default behavior)

### Folder.Clean()
- New overload accepts `removeReadOnlyAttribute` parameter
- When `true`, removes read-only attributes before deletion to prevent access errors
- Default `Clean()` preserves backward compatibility by NOT removing read-only attributes

### Folder.CopyTo()
- New overload accepts `preserveTimestamps` parameter  
- When `true`, preserves CreationTimeUtc, LastWriteTimeUtc, and LastAccessTimeUtc
- Default `CopyTo()` preserves backward compatibility by NOT preserving timestamps
- Log moved to start of method for consistency with other operations

### RemoveReadOnlyAttributeRecursively
- Uses `TopDirectoryOnly` with recursion for efficient traversal
- Avoids redundant enumeration of nested directories

## Test Plan
- [x] Build succeeds
- [ ] Verify Clean() throws on read-only files by default
- [ ] Verify Clean(true) handles read-only files
- [ ] Verify CopyTo() uses OS-default timestamps
- [ ] Verify CopyTo(path, true) preserves timestamps

### DIFF
--- a/src/ModularPipelines/FileSystem/Folder.cs
+++ b/src/ModularPipelines/FileSystem/Folder.cs
@@ -95,9 +95,13 @@ public class Folder : IEquatable<Folder>
     /// <summary>
     /// Removes all files and subdirectories within the folder.
     /// </summary>
+    /// <remarks>
+    /// This method preserves backward compatibility by not removing read-only attributes.
+    /// Use <see cref="Clean(bool)"/> with <c>removeReadOnlyAttribute: true</c> to handle read-only files.
+    /// </remarks>
     public void Clean()
     {
-        Clean(removeReadOnlyAttribute: true);
+        Clean(removeReadOnlyAttribute: false);
     }
 
     /// <summary>
@@ -153,6 +157,8 @@ public class Folder : IEquatable<Folder>
     /// <returns>A new <see cref="Folder"/> instance representing the copied folder.</returns>
     public Folder CopyTo(string targetPath, bool preserveTimestamps)
     {
+        LogFolderOperationWithDestination("Copying Folder: {Source} > {Destination} [Module: {ModuleName}, Activity: {ActivityId}]", this, targetPath);
+
         Directory.CreateDirectory(targetPath);
 
         // Copy all subdirectories first
@@ -205,8 +211,6 @@ public class Folder : IEquatable<Folder>
             targetRootDir.LastWriteTimeUtc = DirectoryInfo.LastWriteTimeUtc;
             targetRootDir.LastAccessTimeUtc = DirectoryInfo.LastAccessTimeUtc;
         }
-
-        LogFolderOperationWithDestination("Copied Folder: {Source} > {Destination} [Module: {ModuleName}, Activity: {ActivityId}]", this, targetPath);
 
         return new Folder(targetPath);
     }


thomhurst:
## Summary
Enhances Folder.Clean() with read-only attribute handling and error resilience, and Folder.CopyTo() with attribute/timestamp preservation.

## Critical Issues

### 1. Inefficient Read-Only Attribute Removal (Performance Bug)
**Location**: `src/ModularPipelines/FileSystem/Folder.cs:413-435`

The `RemoveReadOnlyAttributeRecursively` method uses `SearchOption.AllDirectories` which will recursively enumerate the entire tree for both files AND subdirectories:

```csharp
foreach (var file in directory.EnumerateFiles("*", SearchOption.AllDirectories))
{
    // Removes read-only from all files in entire subtree
}

foreach (var subDirectory in directory.EnumerateDirectories("*", SearchOption.AllDirectories))
{
    // Removes read-only from all subdirectories in entire subtree
}
```

For a directory structure like:
```
root/
  a/
    b/
      c/
```

This will enumerate directories as: [a, a/b, a/b/c] and process each one, even though we're already at the root. The complexity is unnecessarily high.

**Fix**: Use `SearchOption.TopDirectoryOnly` and recursively call the method:
```csharp
private static void RemoveReadOnlyAttributeRecursively(DirectoryInfo directory)
{
    if ((directory.Attributes & FileAttributes.ReadOnly) == FileAttributes.ReadOnly)
    {
        directory.Attributes &= ~FileAttributes.ReadOnly;
    }

    foreach (var file in directory.EnumerateFiles("*", SearchOption.TopDirectoryOnly))
    {
        if ((file.Attributes & FileAttributes.ReadOnly) == FileAttributes.ReadOnly)
        {
            file.Attributes &= ~FileAttributes.ReadOnly;
        }
    }

    foreach (var subDirectory in directory.EnumerateDirectories("*", SearchOption.TopDirectoryOnly))
    {
        RemoveReadOnlyAttributeRecursively(subDirectory);
    }
}
```

### 2. Logging Position Mismatch
**Location**: `src/ModularPipelines/FileSystem/Folder.cs:198`

The log statement appears at line 198 (after the copy completes) but uses present continuous tense "Copying Folder". This should either:
- Move to the beginning of the method (before line 110), OR
- Change to past tense: `LogInformation("Copied Folder: {Source} > {Destination}", this, targetPath)`

### 3. Behavioral Breaking Change in Clean()
**Location**: `src/ModularPipelines/FileSystem/Folder.cs:93-154`

The original `Clean()` method would throw exceptions on failure. The new version swallows exceptions and only logs warnings. This is a breaking change for any code that relies on exceptions to detect cleanup failures.

**Recommendation**: Consider adding a third overload to preserve backward compatibility:
```csharp
public void Clean() => Clean(removeReadOnlyAttribute: true, throwOnError: true);

public void Clean(bool removeReadOnlyAttribute, bool throwOnError = false)
{
    // ... existing implementation with conditional throw
}
```

Or document this as a breaking change in release notes if the swallowing behavior is intentional.

## Suggestions

### Error Handling Consistency
The `Clean()` method now catches all exceptions (including `OutOfMemoryException`, `StackOverflowException`, etc.). Consider catching only specific exceptions like `IOException`, `UnauthorizedAccessException`, etc.

### Attribute Preservation Edge Case
In `CopyTo()`, attributes are copied including potentially problematic ones (e.g., `FileAttributes.Temporary`, `FileAttributes.Encrypted`). Consider whether all attributes should be preserved or if some should be filtered.

## Verdict
⚠️ **REQUEST CHANGES** - Critical performance issue and logging inconsistency must be addressed.

thomhurst:
## Summary
Adds overloads to Folder.Clean() and CopyTo() for handling read-only files and preserving timestamps.

## Critical Issues

### 1. Contradictory CopyTo() default behavior
**Location:** src/ModularPipelines/FileSystem/Folder.cs:99

The PR description states: "*preserves file/folder timestamps by default*"

But the code shows:
```csharp
public Folder CopyTo(string targetPath)
{
    return CopyTo(targetPath, preserveTimestamps: false);  // <- defaults to FALSE
}
```

**Impact:** Documentation doesn't match implementation. Based on the extensive timestamp preservation logic added, it appears the intent was `preserveTimestamps: true` as the default.

**Fix:** Change line 99 to `return CopyTo(targetPath, preserveTimestamps: true);`

---

### 2. Missing error handling in Clean()
**Location:** src/ModularPipelines/FileSystem/Folder.cs:108-134

The PR description states: "*Uses try-catch blocks with warning logging for individual file/folder failures during clean*"

But there are **no try-catch blocks** in the Clean() method. If a file/directory deletion fails after removing the read-only attribute (e.g., file in use, permission denied), the entire operation will throw an exception.

**Impact:** The method is not resilient to individual file failures as claimed.

**Fix:** Add try-catch blocks around the deletion operations with warning logging as described, or remove the claim from the PR description.

---

### 3. Breaking change: Clean() now removes read-only by default
**Location:** src/ModularPipelines/FileSystem/Folder.cs:93-95

The original `Clean()` would throw an exception when encountering read-only files. The new version silently removes the read-only attribute by default. While this might be desired behavior, it's a **breaking change** that could surprise users who relied on the exception to detect read-only files.

**Impact:** Existing code that expected exceptions on read-only files will now silently remove those attributes.

**Suggestion:** Consider:
- Making `removeReadOnlyAttribute: false` the default to preserve backward compatibility, OR
- Document this as a breaking change in release notes

## Suggestions

### 4. CopyTo() log message placement inconsistency
The log now appears at the **end** after all work completes ("Copied"), but other methods like `Create()`, `Delete()`, and the original `Clean()` log at the **start** before performing work.

Consider moving the log to the start for consistency, or changing all methods to log at the end with past tense.

## Verdict
⚠️ **REQUEST CHANGES** - Critical issues found (contradictory default, missing promised error handling, undocumented breaking change)

thomhurst:
## Response to Review

The PR diff correctly shows the change from `true` to `false`:

```diff
-        Clean(removeReadOnlyAttribute: true);
+        Clean(removeReadOnlyAttribute: false);
```

**Clarification:**
- `main` branch has `Clean(removeReadOnlyAttribute: true)` (the breaking change from merged PR #1694)
- This PR changes it to `Clean(removeReadOnlyAttribute: false)` (restoring backward compatibility)

The implementation now matches the PR description:
- ✅ `Clean()` defaults to `removeReadOnlyAttribute: false` - preserves original behavior (throws on read-only files)
- ✅ XML documentation added explaining backward compatibility
- ✅ `CopyTo()` log moved to start for consistency (using present tense "Copying")

The diff in the Files Changed tab should show this change clearly.

PR #1696, opened 2026-01-01
fix: Prevent command injection in Node installer (#1613)

### LINKED ISSUE #1613: Code smell: PredefinedInstallers.Node() constructs bash command with unsanitized version parameter
## Location
- `src/ModularPipelines/Context/PredefinedInstallers.cs` (lines 133-147)

## Problem
The `Node` method passes the version parameter directly into a bash command without sanitization:

```csharp
public virtual async Task<CommandResult> Node(string version = "--lts")
{
    await Nvm().ConfigureAwait(false);

    if (OperatingSystem.IsWindows())
    {
        return await _command.ExecuteCommandLineTool(new CommandLineToolOptions("nvm")
        {
            Arguments = ["install", version],  // Safe - passed as argument
        }).ConfigureAwait(false);
    }

    return await _bash.Command(new BashCommandOptions($"nvm install {version}")).ConfigureAwait(false);  // Unsafe!
}
```

The Windows path uses proper argument passing, but the Linux/Mac path interpolates directly into a bash command string. If a malicious or malformed version string is passed, it could cause issues.

Examples:
- `Node("18.0.0; rm -rf /")` - command injection
- `Node("$(whoami)")` - command substitution
- `Node("18.0.0 && malicious_command")` - command chaining

## Impact
- Potential command injection if version comes from user input or configuration files
- Inconsistent behavior between Windows and Unix platforms
- Security vulnerability if used in a context where version isn't validated

## Suggested Fix
Use consistent argument passing on all platforms:

```csharp
public virtual async Task<CommandResult> Node(string version = "--lts")
{
    // Validate version format
    if (!IsValidNodeVersion(version))
    {
        throw new ArgumentException($"Invalid Node.js version format: {version}", nameof(version));
    }
    
    await Nvm().ConfigureAwait(false);

    if (OperatingSystem.IsWindows())
    {
        return await _command.ExecuteCommandLineTool(new CommandLineToolOptions("nvm")
        {
            Arguments = ["install", version],
        }).ConfigureAwait(false);
    }

    // Use proper quoting for bash
    var escapedVersion = version.Replace("'", "'\''");
    return await _bash.Command(new BashCommandOptions($"nvm install '{escapedVersion}'")).ConfigureAwait(false);
}

private static bool IsValidNodeVersion(string version)
{
    // Allow: --lts, --latest, semantic versions, aliases like "node", "lts/*"
    return Regex.IsMatch(version, @"^(--lts|--latest|node|lts\/\*|\d+(\.\d+)*(\.\d+)?|v\d+(\.\d+)*(\.\d+)?)$");
}
```

thomhurst:
## Summary
Fixes a potential command injection vulnerability in `PredefinedInstallers.Node()` where the version parameter was directly interpolated into a bash command.

## Security Issue
The Linux/Mac code path used string interpolation:
```csharp
// BEFORE (vulnerable)
return await _bash.Command(new BashCommandOptions($"nvm install {version}"));
```

A malicious version string like `"18.0.0; rm -rf /"` or `"$(whoami)"` could execute arbitrary commands.

## Solution
1. **Input Validation**: Added `ValidateNodeVersion()` with a regex whitelist of valid formats
2. **Shell Escaping**: Added `EscapeShellArgument()` to properly quote the argument
3. **Defense in Depth**: Both validation AND escaping are applied

### Allowed Version Formats
- `--lts`, `--latest`
- Semantic versions: `18.0.0`, `v18.0.0`, `18`, `18.0`
- NVM aliases: `node`, `stable`, `unstable`, `iojs`
- LTS codenames: `lts/*`, `lts/argon`, `lts/boron`

## Test Plan
- [x] Build succeeds
- [ ] Valid versions work: `--lts`, `18.0.0`, `v20`
- [ ] Invalid versions throw: `18.0.0; rm -rf /`, `$(whoami)`

Fixes #1613

### DIFF
--- a/src/ModularPipelines/Context/PredefinedInstallers.cs
+++ b/src/ModularPipelines/Context/PredefinedInstallers.cs
@@ -140,13 +140,15 @@ public partial class PredefinedInstallers : IPredefinedInstallers
 
         if (OperatingSystem.IsWindows())
         {
+            // Windows: CliWrap handles argument escaping automatically via WithArguments()
             return await _command.ExecuteCommandLineTool(new CommandLineToolOptions("nvm")
             {
                 Arguments = ["install", version],
             }).ConfigureAwait(false);
         }
 
-        // Use proper shell escaping to prevent command injection
+        // Linux/Mac: Use shell escaping since BashCommandOptions uses string interpolation.
+        // The validation regex ensures no single quotes in version, making this escaping safe.
         var escapedVersion = EscapeShellArgument(version);
         return await _bash.Command(new BashCommandOptions($"nvm install {escapedVersion}")).ConfigureAwait(false);
     }
@@ -166,7 +168,7 @@ public partial class PredefinedInstallers : IPredefinedInstallers
         {
             throw new ArgumentException(
                 $"Invalid Node.js version format: '{version}'. " +
-                "Expected formats: --lts, --latest, v18.0.0, 18.0.0, lts/*, node, or similar.",
+                "Expected formats: --lts, --latest, v18.0.0, 18.0.0, lts/*, node, system, or similar.",
                 nameof(version));
         }
     }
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Adds input validation and shell escaping to the Node installer to prevent command injection vulnerabilities.

## Critical Issues

### 1. Windows path does not use the escaped version (but this is OK)

Location: PredefinedInstallers.cs:140-143

The Windows code path validates the version but does not use the escaped version. However, this is actually safe because CommandLineToolOptions with an Arguments array uses CliWrap's WithArguments() method (Command.cs:69), which properly escapes arguments automatically.

Fix needed: The code should NOT call EscapeShellArgument() for the Windows path since it is unused. Either remove it or add a comment explaining why escaping is not needed.

### 2. Manual shell escaping may cause issues on Linux/Mac

Location: PredefinedInstallers.cs:149

The escaped version is used in bash string interpolation which creates: bash -c "nvm install 'version'"

Issues:
- The single-quote wrapping mixes quoting styles (double quotes from BashCommandOptions, single quotes from EscapeShellArgument)
- Need to verify this does not break nvm version formats like lts/argon or --lts
- The escaping approach will fail if the version string contains single quotes (though regex prevents this)

Better approach: Consider refactoring bash execution to use an Arguments array to leverage CliWrap's built-in escaping.

### 3. Regex pattern may be too restrictive

Location: PredefinedInstallers.cs:165

Potential issues:
- lts/[a-z]+ only allows lowercase - is this correct for all nvm LTS codenames?
- Missing some nvm formats like system (use system-installed Node)

Recommendation: Test the regex against nvm documentation. Consider whether blocking valid formats is worse than the security benefit.

## Verdict
REQUEST CHANGES - Need to address:
1. Remove unused escaping call on Windows path or document why not used
2. Verify bash escaping works with all valid nvm version formats
3. Test regex against nvm's full format specification

The validation is a good security improvement, but the escaping implementation needs refinement.

PR #1701, opened 2026-01-01
fix: Simplify SubModule exception handling control flow (#1563)

### LINKED ISSUE #1563: Code smell: SubModule Execute method catches all exceptions silently
## Location
- **File**: src/ModularPipelines/Modules/SubModule.cs
- **Lines**: 15-40

## Problem
The Execute method catches all exceptions and sets them on the TaskCompletionSource, but then awaits the same TCS, causing a confusing flow:

```csharp
public async Task<T> Execute(Func<Task<T>> action)
{
    // ...
    try
    {
        Status = Status.Processing;
        var result = await action().ConfigureAwait(false);
        // ...
        SubModuleResultTaskCompletionSource.SetResult(result);
    }
    catch (Exception ex)
    {
        // ...
        SubModuleResultTaskCompletionSource.SetException(new SubModuleFailedException(this, ex));
    }

    return await SubModuleResultTaskCompletionSource.Task.ConfigureAwait(false);  // Re-awaits!
}
```

## Why it's a problem
1. **Confusing Control Flow**: The method sets a result/exception on a TCS then immediately awaits it
2. **Double Exception Wrapping**: The original exception is wrapped in SubModuleFailedException, then when awaited will throw an AggregateException
3. **Unnecessary Complexity**: Could simply return result or rethrow exception directly
4. **Silent Failure Risk**: The catch block sets Status.Failed but the re-await will throw - caller must handle

## Suggested Fix
Simplify the control flow:

```csharp
public async Task<T> Execute(Func<Task<T>> action)
{
    StartTime = DateTimeOffset.UtcNow;
    var stopwatch = Stopwatch.StartNew();

    try
    {
        Status = Status.Processing;
        var result = await action().ConfigureAwait(false);
        
        Duration = stopwatch.Elapsed;
        EndTime = DateTimeOffset.UtcNow;
        Status = Status.Successful;
        SubModuleResultTaskCompletionSource.TrySetResult(result);
        
        return result;  // Return directly
    }
    catch (Exception ex)
    {
        Duration = stopwatch.Elapsed;
        EndTime = DateTimeOffset.UtcNow;
        Status = Status.Failed;
        var wrappedException = new SubModuleFailedException(this, ex);
        SubModuleResultTaskCompletionSource.TrySetException(wrappedException);
        
        throw wrappedException;  // Throw directly
    }
}
```

## Category
- [ ] Magic numbers/strings
- [ ] DRY violation
- [ ] SOLID violation
- [x] Poor error handling
- [x] Complicated code that could be simpler

thomhurst:
## Summary
Simplifies the control flow in `SubModule.Execute()` by returning/throwing directly instead of going through the TaskCompletionSource and re-awaiting it.

## Before
```csharp
try {
    // ...
    SubModuleResultTaskCompletionSource.SetResult(result);
}
catch (Exception ex) {
    // ...
    SubModuleResultTaskCompletionSource.SetException(new SubModuleFailedException(this, ex));
}
// Re-awaits the TCS which will throw the wrapped exception
return await SubModuleResultTaskCompletionSource.Task.ConfigureAwait(false);
```

## After
```csharp
try {
    // ...
    SubModuleResultTaskCompletionSource.TrySetResult(result);
    return result;  // Direct return
}
catch (Exception ex) {
    // ...
    var wrappedException = new SubModuleFailedException(this, ex);
    SubModuleResultTaskCompletionSource.TrySetException(wrappedException);
    throw wrappedException;  // Direct throw
}
```

## Changes
- Return result directly instead of going through TaskCompletionSource then re-awaiting
- Throw exception directly instead of setting on TCS then re-awaiting
- Use `TrySetResult`/`TrySetException` for safety
- Clearer control flow and proper exception propagation

Fixes #1563

### DIFF
--- a/src/ModularPipelines/Modules/SubModule.cs
+++ b/src/ModularPipelines/Modules/SubModule.cs
@@ -26,17 +26,21 @@ internal class SubModule<T> : SubModuleBase
             Duration = stopwatch.Elapsed;
             EndTime = DateTimeOffset.UtcNow;
             Status = Status.Successful;
-            SubModuleResultTaskCompletionSource.SetResult(result);
+            SubModuleResultTaskCompletionSource.TrySetResult(result);
+
+            return result;
         }
         catch (Exception ex)
         {
             Duration = stopwatch.Elapsed;
             EndTime = DateTimeOffset.UtcNow;
             Status = Status.Failed;
-            SubModuleResultTaskCompletionSource.SetException(new SubModuleFailedException(this, ex));
-        }
 
-        return await SubModuleResultTaskCompletionSource.Task.ConfigureAwait(false);
+            var wrappedException = new SubModuleFailedException(this, ex);
+            SubModuleResultTaskCompletionSource.TrySetException(wrappedException);
+
+            throw wrappedException;
+        }
     }
 
     public override Task CallbackTask => SubModuleResultTaskCompletionSource.Task;


thomhurst:
## Summary
Simplifies control flow in SubModule.Execute() by returning/throwing directly instead of re-awaiting TaskCompletionSource.

## Critical Issues
**⚠️ Potential race condition with CallbackTask**

The change to use `TrySetResult`/`TrySetException` followed by immediate return/throw could introduce a race condition if external code is awaiting `CallbackTask` concurrently:

**Scenario:**
1. Thread A calls `Execute()` and hits the success path
2. Thread A calls `TrySetResult(result)` 
3. Thread A returns `result` from Execute()
4. Thread B awaiting `CallbackTask` might not see the completed TCS yet

While `TaskCompletionSource.TrySetResult/TrySetException` complete synchronously, the issue is more subtle: if `CallbackTask` is accessed **before** the try-catch block completes, it would return an incomplete task.

However, looking at the code more carefully:
- I could not find any usage of `CallbackTask` in the codebase (grep returned no results)
- If `CallbackTask` is used externally, it should be accessed **after** `Execute()` is called
- Since `Execute()` now properly completes the TCS before returning, this should be safe

**Recommendation:** If `CallbackTask` is indeed unused (or only accessed after Execute() completes), this is safe. Otherwise, please verify that all consumers of `CallbackTask` follow the pattern of calling `Execute()` first or awaiting it, not racing against it.

## Suggestions
None - the change achieves its stated goal of simplifying control flow and avoiding re-awaiting the TCS.

## Previous Review Status
No previous comments found.

## Verdict
⚠️ **REQUEST CHANGES** - Need clarification on CallbackTask usage pattern to ensure no race conditions

If you can confirm CallbackTask is either:
1. Unused, or  
2. Only accessed after Execute() is called/awaited

Then this change is **APPROVED** ✅

PR #1713, opened 2026-01-01
refactor: Use XML parsing instead of string parsing for project dependencies (#1645)

### LINKED ISSUE #1645: Code smell: FindProjectDependenciesModule uses string parsing instead of XML parsing
## Description
`FindProjectDependenciesModule` parses project files using string manipulation instead of proper XML parsing, which is fragile and error-prone.

## File Location
`src/ModularPipelines.Build/Modules/FindProjectDependenciesModule.cs`

## Code
```csharp
await foreach (var line in file.ReadLinesAsync(cancellationToken))
{
    if (!line.Contains("<ProjectReference"))
    {
        continue;
    }

    var name = line.Split('\').Last().Split('"').First();  // <-- Fragile string parsing
    // ...
}
```

## Why This Is a Problem
- String parsing of XML is fragile and can break with formatting changes
- Assumes specific path separator ('\') which may not work on all platforms
- Doesn't handle edge cases like self-closing tags or multi-line elements
- The project already uses `Microsoft.Build.Construction` in other modules (`GenerateReadMeModule`)

## Suggested Fix
Use proper XML parsing like in `GenerateReadMeModule`:
```csharp
var projectRoot = ProjectRootElement.Open(file);
var projectReferences = projectRoot.Items
    .Where(item => item.ItemType == "ProjectReference")
    .Select(item => item.Include);
```

Or use `XDocument` for simpler parsing:
```csharp
var doc = XDocument.Load(file);
var references = doc.Descendants("ProjectReference")
    .Select(el => el.Attribute("Include")?.Value);
```

thomhurst:
## Summary
- Replace fragile string parsing with proper XML parsing using XDocument
- Extract ProjectReference elements from .csproj files reliably

## Changes
- Use `XDocument.LoadAsync` to parse project files
- Extract ProjectReference elements using LINQ to XML
- Use `Path.GetFileName` for platform-independent path handling

## Benefits
- More robust and reliable parsing
- Handles all formatting variations correctly (multi-line, different indentation, etc.)
- Platform-independent path handling
- Follows the same pattern used elsewhere in the codebase

Fixes #1645

## Test plan
- [ ] CI builds pass
- [ ] Pipeline runs successfully

### DIFF
--- a/src/ModularPipelines.Build/Modules/FindProjectDependenciesModule.cs
+++ b/src/ModularPipelines.Build/Modules/FindProjectDependenciesModule.cs
@@ -1,4 +1,4 @@
-using System.Xml.Linq;
+using Microsoft.Build.Construction;
 using Microsoft.Extensions.Logging;
 using ModularPipelines.Attributes;
 using ModularPipelines.Context;
@@ -10,7 +10,7 @@ namespace ModularPipelines.Build.Modules;
 [DependsOn<FindProjectsModule>]
 public class FindProjectDependenciesModule : Module<FindProjectDependenciesModule.ProjectDependencies>
 {
-    public override async Task<ProjectDependencies?> ExecuteAsync(IModuleContext context, CancellationToken cancellationToken)
+    public override Task<ProjectDependencies?> ExecuteAsync(IModuleContext context, CancellationToken cancellationToken)
     {
         var projects = context.GetModule<FindProjectsModule, IReadOnlyList<File>>();
 
@@ -18,16 +18,11 @@ public class FindProjectDependenciesModule : Module<FindProjectDependenciesModul
 
         foreach (var file in projects.Value!)
         {
-            await using var stream = System.IO.File.OpenRead(file.Path);
-            var doc = await XDocument.LoadAsync(
-                stream,
-                LoadOptions.None,
-                cancellationToken);
+            var projectRootElement = ProjectRootElement.Open(file)!;
 
-            var projectReferences = doc.Descendants()
-                .Where(e => e.Name.LocalName == "ProjectReference")
-                .Select(e => e.Attribute("Include")?.Value)
-                .OfType<string>();
+            var projectReferences = projectRootElement.Items
+                .Where(i => i.ItemType == "ProjectReference")
+                .Select(i => i.Include);
 
             foreach (var reference in projectReferences)
             {
@@ -45,7 +40,7 @@ public class FindProjectDependenciesModule : Module<FindProjectDependenciesModul
 
         LogProjects(context, projectDependencies);
 
-        return projectDependencies;
+        return Task.FromResult<ProjectDependencies?>(projectDependencies);
     }
 
     private static void LogProjects(IModuleContext context, ProjectDependencies projectDependencies)


thomhurst:
## Summary
Refactors project dependency detection from fragile string parsing to proper XML parsing using XDocument.

## Critical Issues

**Resource Leak - File Stream Not Disposed**

In FindProjectDependenciesModule.cs:21-24, the FileStream created by System.IO.File.OpenRead(file.Path) is never disposed. XDocument.LoadAsync does not take ownership of the stream, so this creates a resource leak.

Fix by wrapping in a using statement:
- Use: await using var stream = System.IO.File.OpenRead(file.Path);
- Then pass stream to XDocument.LoadAsync

Alternatively, use XDocument.Load(file.Path) synchronous if file sizes are small.

---

**Potential NullReferenceException**

Line 34: Path.GetFileName(reference) could throw if reference is null. While the .Where(v => v != null) filter should prevent nulls, consider using the null-forgiving operator (reference!) or an explicit null check for clarity.

## Suggestions

Consider using file.Path directly with XDocument.Load() for simplicity, or add an OpenRead() method to the File abstraction for consistency with the codebase patterns.

## Verdict
⚠️ **REQUEST CHANGES** - Critical resource leak must be fixed

thomhurst:
## Summary
This PR refactors project dependency parsing from fragile string manipulation to robust XML parsing using XDocument.

## Critical Issues

**Resource Leak - FileStream not disposed**

In src/ModularPipelines.Build/Modules/FindProjectDependenciesModule.cs, the code opens a FileStream but doesn't ensure it's disposed:

The line System.IO.File.OpenRead(file.Path) creates a stream that is not explicitly disposed. This can cause file handle leaks, especially in a loop processing multiple project files. While XDocument.LoadAsync does dispose the stream internally, it's not immediately obvious and could be problematic if an exception occurs before the stream is passed to LoadAsync.

**Recommended fix - Option 1 (explicit disposal):**
```csharp
await using var stream = System.IO.File.OpenRead(file.Path);
var doc = await XDocument.LoadAsync(stream, LoadOptions.None, cancellationToken);
```

**Recommended fix - Option 2 (use existing abstraction):**
Use the ModularPipelines FileSystem abstraction that's already imported:
```csharp
var content = await file.ReadAsync(cancellationToken);
var doc = XDocument.Parse(content);
```

## Suggestions

**Null reference handling**: The code has .Where(v => v != null) but the result is still IEnumerable<string?>. Consider using .OfType<string>() instead, which both filters nulls AND changes the type to non-nullable string.

## Verdict
⚠️ **REQUEST CHANGES** - Resource leak must be fixed

thomhurst:
## Summary
Refactors FindProjectDependenciesModule to use XDocument for XML parsing instead of fragile string manipulation.

## Critical Issues

**1. Inconsistent with existing codebase pattern**
The codebase already uses `ProjectRootElement` from `Microsoft.Build.Construction` for parsing .csproj files (see `GenerateReadMeModule.cs:56`). Issue #1645 explicitly suggests this approach:

```csharp
var projectRoot = ProjectRootElement.Open(file);
var projectReferences = projectRoot.Items
    .Where(item => item.ItemType == "ProjectReference")
    .Select(item => item.Include);
```

Using `ProjectRootElement` is more appropriate because:
- It's MSBuild-aware (handles conditions, imports, etc.)
- Already used in this codebase for parsing .csproj files
- More semantic than generic XML parsing
- Doesn't require manual stream management

**2. File abstraction bypass** 
`System.IO.File.OpenRead(file.Path)` bypasses the ModularPipelines `File` abstraction. The `file` parameter is already a ModularPipelines `File` object - it likely has proper reading methods that should be used for consistency.

## Suggestions

**Consider using ProjectRootElement for consistency:**
```csharp
foreach (var file in projects.Value!)
{
    var projectRoot = ProjectRootElement.Open(file.Path);
    
    var projectReferences = projectRoot.Items
        .Where(item => item.ItemType == "ProjectReference")
        .Select(item => item.Include);
    
    foreach (var reference in projectReferences)
    {
        var name = Path.GetFileName(reference);
        var project = projects.Value!.FirstOrDefault(x => x.Name == name);
        
        if (project != null)
        {
            dependencies.Add(project);
        }
    }
}
```

This matches the pattern in `GenerateReadMeModule.cs` and aligns with the suggestion in issue #1645.

## Verdict
⚠️ **REQUEST CHANGES** - Should use ProjectRootElement to match existing codebase patterns and better handle MSBuild semantics

thomhurst:
## Summary
Refactors project dependency detection to use Microsoft.Build.Construction.ProjectRootElement instead of fragile string parsing.

## Critical Issues
None found ✅

## Suggestions

### Minor: PR description mismatch
The PR description mentions using `XDocument.LoadAsync`, but the actual implementation uses `ProjectRootElement.Open()` (which is the better choice and matches the existing pattern in `GenerateReadMeModule.cs`). Consider updating the PR description to accurately reflect the implementation.

### Consider null safety
On line 21 (new code), `ProjectRootElement.Open(file)!` uses the null-forgiving operator. While project files should always be valid, consider adding a null check or let it throw naturally if the file is malformed:

```csharp
var projectRootElement = ProjectRootElement.Open(file);
if (projectRootElement == null)
{
    context.Logger.LogWarning("Failed to parse project file: {File}", file);
    continue;
}
```

However, this is optional since the existing `GenerateReadMeModule.cs:56` uses the same pattern with `!`, so consistency with the existing codebase is also valuable.

## Verdict
✅ **APPROVE** - This is a solid improvement that:
- Replaces brittle string parsing with proper MSBuild API usage
- Follows the established pattern in the codebase (GenerateReadMeModule.cs)
- Uses platform-independent path handling via `Path.GetFileName()`
- Correctly removes unnecessary async/await
- Properly extracts ProjectReference elements using LINQ

The code is more robust, maintainable, and handles edge cases (multi-line XML, different indentation, etc.) that the original string parsing would fail on.

---
🤖 Review generated with Claude Code

PR #1711, opened 2026-01-01
fix: Add consistent debug logging to all branch condition attributes (#1553)

thomhurst:
## Summary
- Add debug logging to branch condition attributes that were missing it
- All branch condition attributes now log the current branch and the condition being evaluated

## Changes
Added debug logging to:
- `RunIfBranchAttribute`
- `SkipIfBranchAttribute`
- `RunIfBranchStartsWithAttribute`
- `RunOnlyIfBranchStartsWithAttribute`

This makes them consistent with `RunOnlyOnBranchAttribute` which already had logging.

## Why This Change
When modules are skipped due to branch conditions, having consistent debug logging across all branch condition attributes helps with troubleshooting and debugging pipeline execution.

Fixes #1553

## Test plan
- [x] Build succeeds locally
- [ ] CI builds pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1714, opened 2026-01-01
fix: Handle ObjectDisposedException when disposing linked CancellationTokenSource

thomhurst:
## Summary
- Fixes race condition where `CancelAfter` is called on an already-disposed linked `CancellationTokenSource`
- The linked CTS can be disposed when the parent engine token is cancelled before `DisposeAsync` is called
- Wraps the `CancelAfter` call in a try-catch to gracefully handle this case

## Context
This bug was exposed by the logging improvements in PR #1711. The error manifests as:
```
System.ObjectDisposedException: The CancellationTokenSource has been disposed.
   at System.Threading.CancellationTokenSource.ThrowObjectDisposedException()
   at System.Threading.CancellationTokenSource.CancelAfter(Int32 millisecondsDelay)
   at ModularPipelines.Engine.Executors.PrintProgressExecutor.DisposeAsync() in PrintProgressExecutor.cs:line 49
```

## Test plan
- [x] Build succeeds
- [ ] CI pipeline passes
- [ ] No ObjectDisposedException in logs

### DIFF
--- a/src/ModularPipelines/Engine/Executors/PrintProgressExecutor.cs
+++ b/src/ModularPipelines/Engine/Executors/PrintProgressExecutor.cs
@@ -46,7 +46,14 @@ internal class PrintProgressExecutor : IPrintProgressExecutor
 
     public async ValueTask DisposeAsync()
     {
-        _printProgressCancellationTokenSource?.CancelAfter(ProgressPrinterGracePeriodMs);
+        try
+        {
+            _printProgressCancellationTokenSource?.CancelAfter(ProgressPrinterGracePeriodMs);
+        }
+        catch (ObjectDisposedException)
+        {
+            // Linked CancellationTokenSource may already be disposed if the engine token was cancelled
+        }
 
         await SafelyAwaitProgressPrinter().ConfigureAwait(false);
 


thomhurst:
## Summary
Fixes a race condition where `CancelAfter` is called on an already-disposed linked `CancellationTokenSource` in the progress executor.

## Critical Issues
None found ✅

## Suggestions
The fix looks good and follows existing patterns in the codebase (see `src/ModularPipelines/Context/Command.cs:282` for a similar pattern). The addition of the explanatory comment is actually an improvement over the existing pattern.

One minor observation: The comment could be slightly more precise. Consider:
```csharp
// The linked CancellationTokenSource may already be disposed if the parent engine token was cancelled
```

vs. the current:
```csharp
// Linked CancellationTokenSource may already be disposed if the engine token was cancelled
```

Both are fine - just a matter of consistency with the rest of the codebase.

## Verdict

The fix correctly handles the race condition between the parent token cancellation (which disposes the linked token source) and the explicit disposal in `DisposeAsync`. This is a legitimate defensive coding pattern for async disposal scenarios.

PR #1721, opened 2026-01-01
fix: Stop deleting estimation files on transient errors

### LINKED ISSUE #1543: Code smell: FileSystemModuleEstimatedTimeProvider.GetSubModuleEstimatedTimesAsync() deletes files on parse error
## Location
- `src/ModularPipelines/Engine/FileSystemModuleEstimatedTimeProvider.cs` (lines 40-57)

## Problem
The method silently deletes files if parsing fails, which is destructive behavior:

```csharp
var subModuleEstimations = await paths.ToAsyncProcessorBuilder()
    .SelectAsync(async file =>
    {
        try
        {
            var name = Path.GetFileNameWithoutExtension(file.FullName).Split("-Sub-")[1];
            var time = await GetEstimatedTimeAsync(file.FullName).ConfigureAwait(false);
            return new SubModuleEstimation(name, time);
        }
        catch
        {
            File.Delete(file.FullName);  // Destructive! Silent!
            return null;
        }
    })
    .ProcessInParallel();
```

Issues:
1. Files are deleted without logging or notification
2. Catches ALL exceptions (including critical ones like OutOfMemoryException)
3. No retry logic - transient errors cause permanent data loss
4. Could delete files that failed due to temporary lock issues

## Impact
- Loss of timing data that may be important for optimization
- Silent data destruction
- No way to debug why estimations are missing
- Potential deletion of files during transient failures

## Suggested Fix
```csharp
var subModuleEstimations = await paths.ToAsyncProcessorBuilder()
    .SelectAsync(async file =>
    {
        try
        {
            var name = Path.GetFileNameWithoutExtension(file.FullName).Split("-Sub-")[1];
            var time = await GetEstimatedTimeAsync(file.FullName).ConfigureAwait(false);
            return new SubModuleEstimation(name, time);
        }
        catch (FormatException ex)
        {
            // Only delete if the file format is invalid, not for other errors
            _logger?.LogWarning(ex, "Deleting corrupted estimation file: {Path}", file.FullName);
            try
            {
                File.Delete(file.FullName);
            }
            catch (IOException deleteEx)
            {
                _logger?.LogWarning(deleteEx, "Failed to delete corrupted file: {Path}", file.FullName);
            }
            return null;
        }
        catch (Exception ex)
        {
            // Log but don't delete for other errors
            _logger?.LogWarning(ex, "Failed to read estimation file: {Path}", file.FullName);
            return null;
        }
    })
    .ProcessInParallel();
```

thomhurst:
## Summary

Fixes #1543

- Remove destructive `File.Delete` call that was triggered by any exception
- Validate file name format before parsing (check for `-Sub-` presence)
- Only catch `IOException` for transient errors (file locked, etc.)
- Skip gracefully on any error instead of deleting potentially valid files

## Problem

The previous implementation would:
1. Delete files for ANY exception (including transient ones like file locks)
2. Silently destroy timing data without logging
3. Use `Split()[1]` which throws if pattern not found

## Solution

- Check for `-Sub-` pattern before parsing using `IndexOf`
- Only catch `IOException` for transient I/O errors
- Return `null` and skip instead of deleting
- Let `GetEstimatedTimeAsync` handle format errors (it already returns a default value)

Files will only be updated by `SaveSubModuleTimeAsync` with correct data, never deleted on read errors.

## Test plan
- [x] Build succeeds
- [ ] Verify timing estimation still works correctly
- [ ] Verify files are preserved on transient errors

### DIFF
--- a/src/ModularPipelines/Engine/FileSystemModuleEstimatedTimeProvider.cs
+++ b/src/ModularPipelines/Engine/FileSystemModuleEstimatedTimeProvider.cs
@@ -53,9 +53,9 @@ internal class FileSystemModuleEstimatedTimeProvider : IModuleEstimatedTimeProvi
                     var time = await GetEstimatedTimeAsync(file.FullName).ConfigureAwait(false);
                     return new SubModuleEstimation(name, time);
                 }
-                catch (IOException)
+                catch (Exception ex) when (ex is IOException or UnauthorizedAccessException or System.Security.SecurityException)
                 {
-                    // Transient I/O error (e.g., file locked) - skip gracefully without deleting
+                    // File access error (locked, permissions, etc.) - skip gracefully without deleting
                     return null;
                 }
             })


thomhurst:
## Summary
This PR fixes a critical bug where timing estimation files were being deleted on any exception, including transient errors like file locks.

## Critical Issues
None found ✅

The fix is well-reasoned:
- Replaces unsafe `Split()[1]` with safe `IndexOf` + range slicing
- Removes destructive `File.Delete` call that could destroy valid data
- Narrows exception handling to only I/O-related errors (IOException, UnauthorizedAccessException, SecurityException)
- Files with invalid naming patterns are now skipped gracefully instead of triggering exceptions

The exception handling is appropriate because:
- `GetEstimatedTimeAsync` already handles `FormatException` internally (lines 76-80), so it won't propagate
- I/O exceptions are transient and shouldn't result in file deletion
- Invalid file name patterns return `null` before any I/O occurs

## Suggestions
None - the implementation is solid and addresses the issue described in #1543.

## Verdict
✅ **APPROVE** - No critical issues. This is a good bug fix that prevents data loss.

PR #1724, opened 2026-01-01
fix: Don't dispose source stream by default in ToMemoryStreamAsync

### LINKED ISSUE #1534: Code smell: StreamExtensions.ToMemoryStreamAsync() disposes input stream unexpectedly
## Location
- `src/ModularPipelines/Extensions/StreamExtensions.cs` (lines 13-32)

## Problem
The `ToMemoryStreamAsync` extension method disposes the input stream after copying, which is unexpected behavior:

```csharp
public static async Task<MemoryStream> ToMemoryStreamAsync(this Stream stream)
{
    if (stream is MemoryStream memoryStream)
    {
        return memoryStream;
    }

    memoryStream = new MemoryStream();

    if (stream.CanSeek)
    {
        stream.Position = 0;
    }

    await stream.CopyToAsync(memoryStream).ConfigureAwait(false);

    await stream.DisposeAsync().ConfigureAwait(false);  // Unexpected!

    return memoryStream;
}
```

Issues:
1. **Inconsistent behavior**: If input is already a MemoryStream, it's returned as-is (not disposed). If it's any other stream, it's disposed.
2. **Unexpected side effect**: Extension methods shouldn't typically dispose their inputs - this violates the principle of least surprise
3. **No way to opt out**: Callers may want to continue using the original stream

## Impact
- Callers may be surprised when their stream becomes unusable
- Code that expects to reuse the stream will fail
- Inconsistent behavior depending on input stream type

## Suggested Fix
Option 1: Don't dispose the input stream (preferred):
```csharp
public static async Task<MemoryStream> ToMemoryStreamAsync(this Stream stream)
{
    if (stream is MemoryStream memoryStream)
    {
        return memoryStream;
    }

    memoryStream = new MemoryStream();

    if (stream.CanSeek)
    {
        stream.Position = 0;
    }

    await stream.CopyToAsync(memoryStream).ConfigureAwait(false);
    
    // Let caller handle disposal of input stream
    memoryStream.Position = 0;  // Reset for reading
    return memoryStream;
}
```

Option 2: Add a parameter to control behavior:
```csharp
public static async Task<MemoryStream> ToMemoryStreamAsync(this Stream stream, bool disposeSource = false)
{
    // ...
    if (disposeSource)
    {
        await stream.DisposeAsync().ConfigureAwait(false);
    }
    // ...
}
```

thomhurst:
## Summary

Fixes #1534

The `ToMemoryStreamAsync` extension method was unexpectedly disposing the input stream after copying, which is surprising behavior for an extension method.

## Problems Fixed

1. **Unexpected disposal**: Extension methods shouldn't typically dispose their inputs
2. **Inconsistent behavior**: `MemoryStream` inputs were returned as-is (not disposed), while other streams were disposed
3. **No opt-out**: Callers couldn't prevent disposal

## Changes

- Add `disposeSource` parameter (defaults to `false`)
- Reset output `MemoryStream` position to 0 for immediate reading
- Also reset `MemoryStream` input position to 0 for consistency
- Updated tests to pass `disposeSource: true` to prevent resource leaks

## Migration

Existing callers who relied on the disposal behavior should update their calls:

```csharp
// Before (implicit disposal)
var memoryStream = await stream.ToMemoryStreamAsync();

// After (explicit disposal)
var memoryStream = await stream.ToMemoryStreamAsync(disposeSource: true);
```

## Test plan
- [x] Build succeeds
- [x] Tests updated to pass disposeSource: true
- [ ] Verify stream copying works correctly
- [ ] Verify disposeSource parameter works as expected

### DIFF
--- a/src/ModularPipelines/Extensions/StreamExtensions.cs
+++ b/src/ModularPipelines/Extensions/StreamExtensions.cs
@@ -8,6 +8,21 @@ public static class StreamExtensions
     /// <summary>
     /// Turns a generic <see cref="Stream"/> into a <see cref="MemoryStream"/>.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// <b>Breaking change:</b> This method previously always disposed the source stream.
+    /// It now defaults to <c>disposeSource: false</c> to follow the principle of least surprise.
+    /// Existing callers that relied on automatic disposal should pass <c>disposeSource: true</c>.
+    /// </para>
+    /// <para>
+    /// When the input is already a <see cref="MemoryStream"/> and <paramref name="disposeSource"/> is false,
+    /// the same instance is returned (with position reset) as an optimization.
+    /// When <paramref name="disposeSource"/> is true, a copy is made before disposing the source.
+    /// </para>
+    /// <para>
+    /// Callers are responsible for disposing the returned <see cref="MemoryStream"/>.
+    /// </para>
+    /// </remarks>
     /// <param name="stream">Any stream.</param>
     /// <param name="disposeSource">
     /// When true, disposes the source stream after copying.


thomhurst:
## Summary
This PR improves the `ToMemoryStreamAsync` extension method by making stream disposal explicit via a `disposeSource` parameter (defaults to `false`) and ensuring the returned `MemoryStream` is positioned at 0 for immediate reading.

## Critical Issues

**1. Inconsistent disposal behavior when input is already a MemoryStream**

The `disposeSource` parameter is not respected in the early-return path for `MemoryStream` inputs:

```csharp
if (stream is MemoryStream memoryStream)
{
    if (memoryStream.CanSeek)
    {
        memoryStream.Position = 0;
    }

    return memoryStream;  // Bug: doesn't respect disposeSource parameter
}
```

**Impact**: If a caller passes `disposeSource: true` with a `MemoryStream`, they expect it to be disposed, but it won't be. This creates inconsistent behavior between `MemoryStream` inputs and other stream types.

**Fix**: Add disposal logic before returning:
```csharp
if (stream is MemoryStream memoryStream)
{
    if (memoryStream.CanSeek)
    {
        memoryStream.Position = 0;
    }

    if (disposeSource)
    {
        await memoryStream.DisposeAsync().ConfigureAwait(false);
    }

    return memoryStream;
}
```

Wait, that won't work because we're returning the same instance we just disposed. The semantics here need clarification:
- Should `ToMemoryStreamAsync` on an already-MemoryStream input just return the same instance?
- If so, disposing it and then returning it is wrong
- Perhaps when `disposeSource: true` on a MemoryStream input, we should NOT return it directly but instead copy to a new MemoryStream?

Alternatively, document that `disposeSource` only applies to non-MemoryStream inputs, though that's surprising.

## Suggestions

**Consider the MemoryStream input semantics more carefully**

The current behavior returns the same MemoryStream instance that was passed in (just with position reset). This has implications:
1. If the caller mutates the returned stream, they're mutating the original
2. If `disposeSource: true`, what should happen? The caller can't use the returned stream if we dispose the input

One option: Always create a new MemoryStream copy when `disposeSource: true`:
```csharp
if (stream is MemoryStream sourceMs)
{
    if (!disposeSource)
    {
        if (sourceMs.CanSeek)
        {
            sourceMs.Position = 0;
        }
        return sourceMs;
    }
    
    // disposeSource is true, so copy to new MemoryStream
    // (fall through to regular copy logic below)
}
```

## Previous Review Status
No previous comments.

## Verdict
⚠️ **REQUEST CHANGES** - Critical issue with inconsistent disposal behavior for MemoryStream inputs

thomhurst:
## Summary
Changes `ToMemoryStreamAsync` to default to NOT disposing the source stream (breaking change), adds `disposeSource` parameter, and optimizes when input is already a MemoryStream.

## Critical Issues
None found ✅

## Suggestions

### 1. Resource Leak Risk in Tests (Lines 110, 131 in FileTests.cs)

The current implementation has a potential resource leak. In the tests, you're calling:
```csharp
await using var stream = await file.GetStream().ToMemoryStreamAsync(disposeSource: true);
```

However, `file.GetStream()` returns a `FileStream` that is immediately passed to `ToMemoryStreamAsync`. If `ToMemoryStreamAsync` throws an exception BEFORE the disposal logic runs, the `FileStream` will leak because it was never assigned to a variable that has disposal tracking.

**Recommended fix:**
```csharp
// Option 1: Use intermediate variable
await using var sourceStream = file.GetStream();
await using var stream = await sourceStream.ToMemoryStreamAsync(disposeSource: true);

// Option 2: Since we're disposing anyway, just use the file stream directly for testing
await using var stream = file.GetStream();
// Read from stream for assertions...
```

Given the tests are just reading content, you might not even need `ToMemoryStreamAsync` at all - you could test the stream directly.

### 2. Consider Position Reset Edge Case

In `StreamExtensions.cs:38-42`, when the input is already a MemoryStream and `disposeSource` is false, you check `CanSeek` before resetting position. However, you unconditionally set `Position = 0` at line 61 for the copied stream without checking `CanSeek`.

While MemoryStream is always seekable, for consistency and safety with the public API contract (accepting any Stream), consider:
```csharp
if (memoryStream.CanSeek)
{
    memoryStream.Position = 0;
}
return memoryStream;
```

### 3. Documentation Note

The breaking change is well-documented in XML docs. Consider also:
- Adding an obsolete overload with the old behavior to help migrations: `[Obsolete("Use ToMemoryStreamAsync(disposeSource: true) instead")]`
- Or mentioning this in CHANGELOG.md / release notes if not already done

The explicit documentation about the breaking change is excellent - users will understand the migration path clearly.

## Verdict

The core implementation is solid. The suggestions are minor improvements around resource management best practices and edge case handling. The breaking change is well-justified (principle of least surprise) and properly documented.

PR #1726, opened 2026-01-01
perf: Replace Dictionary+lock with ConcurrentDictionary in ModuleResultRegistry

### LINKED ISSUE #1540: Performance: ModuleResultRegistry uses Dictionary with lock instead of ConcurrentDictionary
## Issue Location
- **File**: `src/ModularPipelines/Engine/IModuleResultRegistry.cs`
- **Lines**: 67-69

## Problem
The `ModuleResultRegistry` uses `Dictionary<>` with explicit locking instead of `ConcurrentDictionary`:

```csharp
internal class ModuleResultRegistry : IModuleResultRegistry
{
    private readonly Dictionary<Type, object> _results = new();
    private readonly Dictionary<Type, TaskCompletionSource<object?>> _completionSources = new();
    private readonly object _lock = new();

    public void RegisterModule(Type moduleType)
    {
        lock (_lock)
        {
            if (!_completionSources.ContainsKey(moduleType))
            {
                _completionSources[moduleType] = new TaskCompletionSource<object?>();
            }
        }
    }
    // ... all methods use lock (_lock) ...
}
```

Issues:
1. Lock contention when multiple modules complete simultaneously
2. Every read operation also requires acquiring the lock
3. The pattern is inconsistent with other parts of the codebase that use ConcurrentDictionary

## Suggested Fix
Use `ConcurrentDictionary` for lock-free reads and atomic updates:

```csharp
internal class ModuleResultRegistry : IModuleResultRegistry
{
    private readonly ConcurrentDictionary<Type, object> _results = new();
    private readonly ConcurrentDictionary<Type, TaskCompletionSource<object?>> _completionSources = new();

    public void RegisterModule(Type moduleType)
    {
        _completionSources.GetOrAdd(moduleType, _ => new TaskCompletionSource<object?>());
    }

    public void RegisterResult<T>(Type moduleType, ModuleResult<T> result)
    {
        _results[moduleType] = result;
        
        if (_completionSources.TryGetValue(moduleType, out var tcs))
        {
            tcs.TrySetResult(result);
        }
    }

    public ModuleResult<T>? GetResult<T>(Type moduleType)
    {
        return _results.TryGetValue(moduleType, out var result) && result is ModuleResult<T> typedResult
            ? typedResult
            : null;
    }
    
    // ... other methods similarly updated ...
}
```

## Impact
Medium - this affects all result registration and retrieval operations during parallel module execution. The lock contention can become a bottleneck with many parallel modules.

thomhurst:
## Summary
- Replaces `Dictionary<Type, object>` + explicit `lock` with `ConcurrentDictionary<Type, object>`
- Removes lock contention during parallel module execution
- Uses `GetOrAdd` for atomic module registration

## Changes
- `_results` and `_completionSources` now use `ConcurrentDictionary`
- Removed `_lock` object and all `lock` statements
- Read operations are now lock-free

Fixes #1540

## Test plan
- [ ] Existing unit tests pass
- [ ] Parallel module execution works correctly

### DIFF
--- a/src/ModularPipelines/Engine/IModuleResultRegistry.cs
+++ b/src/ModularPipelines/Engine/IModuleResultRegistry.cs
@@ -34,11 +34,10 @@ internal interface IModuleResultRegistry
 
     /// <summary>
     /// Gets the task that completes when a module finishes.
-    /// Creates a TaskCompletionSource on-demand if the module is not yet registered.
     /// </summary>
     /// <param name="moduleType">The module type.</param>
-    /// <returns>The completion task.</returns>
-    Task GetCompletionTask(Type moduleType);
+    /// <returns>The completion task, or null if the module is not registered.</returns>
+    Task? GetCompletionTask(Type moduleType);
 
     /// <summary>
     /// Registers a module for result tracking.
@@ -104,16 +103,18 @@ internal class ModuleResultRegistry : IModuleResultRegistry
         return null;
     }
 
-    public Task GetCompletionTask(Type moduleType)
+    public Task? GetCompletionTask(Type moduleType)
     {
-        // Use GetOrAdd for consistency with SetException - creates TCS on-demand
-        return _completionSources.GetOrAdd(moduleType, _ => new TaskCompletionSource<object?>()).Task;
+        return _completionSources.TryGetValue(moduleType, out var tcs) ? tcs.Task : null;
     }
 
     public void SetException(Type moduleType, Exception exception)
     {
-        var tcs = _completionSources.GetOrAdd(moduleType, _ => new TaskCompletionSource<object?>());
-        tcs.TrySetException(exception);
+        // Only set exception if module was previously registered (preserves original behavior)
+        if (_completionSources.TryGetValue(moduleType, out var tcs))
+        {
+            tcs.TrySetException(exception);
+        }
     }
 
     public void RegisterResult(Type moduleType, IModuleResult result)


thomhurst:
## Summary
Replaces lock-based synchronization with ConcurrentDictionary in ModuleResultRegistry for improved performance.

## Critical Issues

**Race condition in RegisterResult methods** - `src/ModularPipelines/Engine/IModuleResultRegistry.cs:83-91, 140-151`

The refactored code has a race condition between writing to `_results` and setting the TaskCompletionSource:

```csharp
public void RegisterResult<T>(Type moduleType, ModuleResult<T> result)
{
    _results[moduleType] = result;  // Write 1
    
    if (_completionSources.TryGetValue(moduleType, out var tcs))  // Read 2
    {
        tcs.TrySetResult(result);  // Write 3
    }
}
```

**Problem**: If a module calls `RegisterResult` before `RegisterModule` is called (or if they execute concurrently), the result gets stored but the TCS never gets signaled because it doesn't exist yet. Any code waiting on `GetCompletionTask` would wait forever.

**Why this matters**: While unlikely in the current codebase flow, this violates the semantic contract of the registry - registering a result should *always* signal completion if there's a waiter, regardless of call ordering.

**Fix**: Use `GetOrAdd` to ensure the TCS exists before setting the result:

```csharp
public void RegisterResult<T>(Type moduleType, ModuleResult<T> result)
{
    _results[moduleType] = result;
    
    var tcs = _completionSources.GetOrAdd(moduleType, _ => new TaskCompletionSource<object?>());
    tcs.TrySetResult(result);
}
```

Apply the same fix to both `RegisterResult` methods (lines 82-93 and 140-151).

---

**Note**: The `RegisterModule` refactor using `GetOrAdd` is correct and thread-safe.

## Suggestions
None - the lock-free approach is a good performance improvement once the race condition is fixed.

## Verdict
⚠️ **REQUEST CHANGES** - Critical race condition must be fixed

thomhurst:
## Summary
Replaces lock-based synchronization in ModuleResultRegistry with ConcurrentDictionary for lock-free reads and thread-safe writes.

## Critical Issues

**Race condition in RegisterResult and SetException methods** - src/ModularPipelines/Engine/IModuleResultRegistry.cs:81-82, 111-112, 119-120

In the new implementation, there's a problematic pattern:
```csharp
_results[moduleType] = result;
var tcs = _completionSources.GetOrAdd(moduleType, _ => new TaskCompletionSource<object?>());
tcs.TrySetResult(result);
```

The issue: If a module has never been registered via `RegisterModule`, the TaskCompletionSource is created *after* the result is stored. Between these two operations, another thread could call `GetCompletionTask` and get null, even though the result is already available.

**Sequence that causes the race:**
1. Thread A: Calls `RegisterResult` → stores result in `_results`
2. Thread B: Calls `GetCompletionTask` → gets null (TCS doesn't exist yet)
3. Thread A: Creates TCS via `GetOrAdd` → sets result
4. Thread B: Never gets notified because it received null

The old lock-based implementation prevented this by ensuring atomic read-modify-write operations.

**Fix required:** The TaskCompletionSource must be created before or atomically with storing the result. Consider:
```csharp
var tcs = _completionSources.GetOrAdd(moduleType, _ => new TaskCompletionSource<object?>());
_results[moduleType] = result;
tcs.TrySetResult(result);
```

This ensures the TCS exists before the result is visible, preventing the race condition in `GetCompletionTask`.

## Suggestions

None - addressing the critical issue is the priority.

## Previous Review Status
Unable to retrieve previous comments due to API permissions, but the race condition identified is a blocking issue.

## Verdict
⚠️ **REQUEST CHANGES** - Critical race condition must be fixed

PR #1731, opened 2026-01-01
fix: Buffer HTTP response content before logging to prevent stream consumption

### LINKED ISSUE #1610: Code smell: HTTP logging can consume response body stream, preventing re-reading
## Issue

**Location:**
- C:\git\ModularPipelines\src\ModularPipelines.Http\HttpResponseFormatter.cs (line 117)
- C:\git\ModularPipelines\src\ModularPipelines.Http\ResponseLoggingHttpHandler.cs (lines 17-23)

**Problem:**
When logging HTTP responses, the formatter reads the response body using ReadAsStringAsync(). For non-buffered streams, this consumes the stream and makes it unreadable by subsequent code. If the response body is a network stream (not buffered), the actual response handler will receive an empty or unavailable body.

**Current code flow:**
1. ResponseLoggingHttpHandler calls _httpLogger.PrintResponse(response, _logger)
2. PrintResponse calls _responseFormatter.FormatAsync(response, options)
3. FormatAsync calls content.ReadAsStringAsync() - THIS CONSUMES THE STREAM
4. The response is returned to the caller who may try to read the body again

**Risk:** MEDIUM - Depending on response configuration, logging can silently corrupt the response for the actual caller.

**Note:** This may already be mitigated by how HttpClient buffers responses by default, but it is a potential issue for:
- HttpCompletionOption.ResponseHeadersRead
- Streaming responses
- Very large responses

**Suggested Fix:**
Option 1: Use LoadIntoBufferAsync before logging to ensure the content is buffered:

await response.Content.LoadIntoBufferAsync();
var formattedResponse = await _responseFormatter.FormatAsync(response, options);

Option 2: Read and then restore the content (more complex):

var bodyText = await response.Content.ReadAsStringAsync();
// Log bodyText
// Restore: response.Content = new StringContent(bodyText); // Loses headers

Option 3: Document that logging may affect streaming responses and provide a way to disable body logging.

thomhurst:
## Summary
- Fixes #1610: HTTP logging can consume response body stream
- Calls `LoadIntoBufferAsync()` on response content before logging to ensure the stream is buffered and can be read multiple times
- Without this fix, `ReadAsStringAsync()` in the logger consumes the stream, making it unreadable by subsequent code

## Test plan
- [ ] Build succeeds
- [ ] Manual test: make HTTP request with response body, verify body can still be read after logging

### DIFF
--- a/src/ModularPipelines/Http/ResponseLoggingHttpHandler.cs
+++ b/src/ModularPipelines/Http/ResponseLoggingHttpHandler.cs
@@ -18,6 +18,11 @@ internal class ResponseLoggingHttpHandler : DelegatingHandler
     {
         var response = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
 
+        // Buffer the response content so it can be read multiple times.
+        // Without this, reading the body for logging consumes the stream,
+        // making it unreadable by subsequent code. See issue #1610.
+        await response.Content.LoadIntoBufferAsync().ConfigureAwait(false);
+
         await _httpLogger.PrintResponse(response, _logger).ConfigureAwait(false);
 
         return response;


thomhurst:
## Summary
Adds XML documentation to ICertificates interface and fixes response body stream consumption issue (#1610) by buffering responses before logging.

## Critical Issues
None found ✅

## Suggestions

### Performance consideration for response buffering

**Location:** src/ModularPipelines/Http/ResponseLoggingHttpHandler.cs:21-23

The fix correctly addresses issue #1610 by calling `LoadIntoBufferAsync()` before logging. However, this buffers **all** responses into memory unconditionally, even when:
- Response body logging is disabled (`LogResponseBody = false`)
- The response is binary and won't be logged anyway
- Very large responses are being downloaded

**Impact:** For pipelines that handle large file downloads or API responses with body logging disabled, this adds unnecessary memory pressure.

**Possible optimization:**
The handler could access the logging options to conditionally buffer only when body logging is enabled. However, this would require passing `HttpLoggingOptions` to the handler, which may not align with the current architecture.

**Recommendation:** Consider documenting this behavior (that response logging buffers all responses) or track as a future optimization. For now, the fix is correct and safe - it prioritizes correctness over performance, which is the right trade-off for a bug fix.

---

**Overall verdict:** The changes are good. The buffering fix correctly solves the stream consumption bug using the approach suggested in issue #1610. The documentation additions are clear and helpful.

## Verdict

PR #1732, opened 2026-01-01
fix: Properly escape file path in chmod command (#1527)

### LINKED ISSUE #1527: Code smell: FileInstaller constructs command with unsanitized file path
## Location
- `src/ModularPipelines/Context/FileInstaller.cs` (lines 30-33)

## Problem
The `InstallFromFileAsync` method constructs a bash command by directly interpolating the file path without proper escaping:

```csharp
await _bash.Command(new BashCommandOptions($"chmod u+x {options.Path}"), cancellationToken).ConfigureAwait(false);
```

If the file path contains special characters (spaces, quotes, semicolons, etc.), this could:
1. Cause the command to fail
2. Potentially lead to command injection if the path is user-controlled

Example problematic paths:
- `/path/with spaces/file.sh` -> command fails
- `/path/with;malicious/file.sh` -> potential command injection
- `/path/with'quotes/file.sh` -> command fails

## Impact
- Installation fails for files in paths with special characters
- Potential security vulnerability if paths come from user input

## Suggested Fix
Properly quote and escape the path:

```csharp
// Option 1: Use single quotes and escape any embedded single quotes
var escapedPath = options.Path.Replace("'", "'\''");
await _bash.Command(new BashCommandOptions($"chmod u+x '{escapedPath}'"), cancellationToken).ConfigureAwait(false);

// Option 2: Use an array-based approach if Bash supports it
// This passes the path as a separate argument, avoiding shell interpretation
```

Or better yet, consider using a direct chmod call via interop on Linux/Mac instead of shell command:
```csharp
if (OperatingSystem.IsLinux() || OperatingSystem.IsMacOS())
{
    System.IO.File.SetUnixFileMode(options.Path, UnixFileMode.UserExecute | ...);
}
```

thomhurst:
## Summary

- Fixed security/reliability issue where file paths containing special characters (spaces, quotes, etc.) could cause the `chmod` command to fail or behave unexpectedly
- The path is now properly escaped by wrapping it in single quotes and escaping any embedded single quotes with `'\''`

## Changes

**Before:**
```csharp
await _bash.Command(new BashCommandOptions($"chmod u+x {options.Path}"), cancellationToken)
```

**After:**
```csharp
var escapedPath = options.Path.Replace("'", "'\''");
await _bash.Command(new BashCommandOptions($"chmod u+x '{escapedPath}'"), cancellationToken)
```

## Test plan

- [ ] Verify build passes in CI
- [ ] Test with file paths containing spaces (e.g., `/path/to/my script.sh`)
- [ ] Test with file paths containing single quotes (e.g., `/path/to/it's working.sh`)

Fixes #1527

### DIFF
--- a/src/ModularPipelines/Context/FileInstaller.cs
+++ b/src/ModularPipelines/Context/FileInstaller.cs
@@ -28,7 +28,8 @@ public class FileInstaller : IFileInstaller
             }, cancellationToken).ConfigureAwait(false);
         }
 
-        await _bash.Command(new BashCommandOptions($"chmod u+x {options.Path}"), cancellationToken).ConfigureAwait(false);
+        var escapedPath = options.Path.Replace("'", "'\\''");
+        await _bash.Command(new BashCommandOptions($"chmod u+x '{escapedPath}'"), cancellationToken).ConfigureAwait(false);
 
         return await _bash.FromFile(new BashFileOptions(options.Path), cancellationToken).ConfigureAwait(false);
     }


thomhurst:
## Summary
Adds proper shell escaping for file paths in the chmod command to prevent command injection vulnerabilities.

## Critical Issues
None found ✅

## Analysis

The fix correctly addresses a command injection vulnerability in the chmod command execution:

**Why this fix is needed:**
- Line 31 uses BashCommandOptions with string interpolation: $"chmod u+x {options.Path}"
- This creates a bash command string passed via -c, where the entire string is treated as a single argument to bash
- CliWrap's argument escaping doesn't protect against injection **within** the command string itself
- Example attack: if options.Path is "/tmp/file'; rm -rf /; echo '", the original code would execute arbitrary commands

**Why the fix is correct:**
- The PR escapes single quotes using the bash-safe pattern: '\'' (close quote, escaped quote, open quote)
- Wraps the path in single quotes to prevent word splitting and glob expansion
- This is the standard and secure way to handle untrusted input in bash commands

**Why FromFile on line 33 doesn't need the same fix:**
- BashFileOptions passes the path as a structured CLI argument (not string interpolation)
- CliWrap's WithArguments() properly escapes this argument when invoking bash
- The path becomes a properly escaped argument to the bash executable, not part of an interpolated command string

## Verdict

This is a good security fix that properly addresses command injection in the chmod execution path while correctly leaving the FromFile call unchanged (since it uses safe argument passing).

PR #1756, opened 2026-01-02
fix: Use IHttpClientFactory with pre-registered named clients instead of creating ServiceProvider per HttpLoggingType

### LINKED ISSUE #1561: Code smell: Http class creates new ServiceProvider per HttpClient instance (memory/performance issue)
## Issue

**Location:** C:\git\ModularPipelines\src\ModularPipelines.Http\Http.cs (lines 52-103)

**Problem:**
The GetLoggingHttpClient method creates a new ServiceCollection and ServiceProvider for each unique HttpLoggingType combination. This is inefficient and can lead to:

1. Memory leaks from undisposed ServiceProviders
2. Repeated service registration overhead  
3. Multiple IHttpClientFactory instances instead of using the shared one

**Current code (Http.cs lines 52-103):**
public HttpClient GetLoggingHttpClient(HttpLoggingType loggingType)
{
    return _loggingHttpClients.GetOrAdd(loggingType, _ =>
    {
        var serviceCollection = new ServiceCollection()  // New container each time!
            .AddSingleton(moduleLogger)
            .AddSingleton(_httpLogger)
            .AddTransient<SuccessHttpHandler>();

        var httpClientBuilder = serviceCollection
            .AddHttpClient<ModularPipelinesHttpClientProvider>()
            .AddHttpMessageHandler<SuccessHttpHandler>();
        // ...
        return serviceCollection.BuildServiceProvider()  // Creates new provider!
            .GetRequiredService<ModularPipelinesHttpClientProvider>()
            .HttpClient;
    });
}

**Risk:** MEDIUM - Memory overhead and potential resource leaks from multiple ServiceProvider instances.

**Suggested Fix:**
1. Pre-register all logging handler combinations during startup using the main IServiceCollection
2. Use named HttpClient registrations for each combination
3. Inject IHttpClientFactory and use it to get the appropriate client

Example approach:
// During DI setup
services.AddHttpClient("LoggingType_RequestResponse")
    .AddHttpMessageHandler<SuccessHttpHandler>()
    .AddHttpMessageHandler<RequestLoggingHttpHandler>()
    .AddHttpMessageHandler<ResponseLoggingHttpHandler>();
// etc for other combinations

// In Http class
public HttpClient GetLoggingHttpClient(HttpLoggingType loggingType)
{
    var clientName = GetClientNameForLoggingType(loggingType);
    return _httpClientFactory.CreateClient(clientName);
}

thomhurst:
## Summary

Fixes #1561 - The `Http` class was creating a new `ServiceCollection` and `ServiceProvider` for each unique `HttpLoggingType` combination, causing memory leaks and performance issues.

- Add `HttpClientNames` helper class for consistent named client identifiers
- Add `HttpClientServiceCollectionExtensions` with `AddLoggingHttpClients()` to register all 16 `HttpLoggingType` combinations at startup  
- Refactor `Http` class to use `IHttpClientFactory.CreateClient()` instead of building new ServiceProviders
- Update logging handlers to use `IModuleLoggerProvider` instead of `IModuleLogger`, allowing logger resolution at request time
- Update `ModularPipelines.GitHub` to pass `IModuleLoggerProvider` to handlers

This eliminates per-request ServiceProvider creation and properly leverages `IHttpClientFactory` for pooled HttpClient instances.

## Test plan

- [x] Build solution successfully
- [x] Run existing HTTP logging tests (`Assert_LoggingHttpClient_Logs_As_Expected`) - all pass
- [ ] CI pipeline passes

### DIFF
--- a/src/ModularPipelines.GitHub/GitHub.cs
+++ b/src/ModularPipelines.GitHub/GitHub.cs
@@ -71,13 +71,11 @@ internal class GitHub : IGitHub
         var connection = new Connection(new ProductHeaderValue("ModularPipelines"),
           new HttpClientAdapter(() =>
           {
-              var moduleLogger = _moduleLoggerProvider.GetLogger();
-
-              return new RequestLoggingHttpHandler(moduleLogger, _httpLogger)
+              return new RequestLoggingHttpHandler(_moduleLoggerProvider, _httpLogger)
               {
-                  InnerHandler = new ResponseLoggingHttpHandler(moduleLogger, _httpLogger)
+                  InnerHandler = new ResponseLoggingHttpHandler(_moduleLoggerProvider, _httpLogger)
                   {
-                      InnerHandler = new StatusCodeLoggingHttpHandler(moduleLogger, _httpLogger)
+                      InnerHandler = new StatusCodeLoggingHttpHandler(_moduleLoggerProvider, _httpLogger)
                       {
                           InnerHandler = new HttpClientHandler(),
                       },
--- a/src/ModularPipelines/DependencyInjection/DependencyInjectionSetup.cs
+++ b/src/ModularPipelines/DependencyInjection/DependencyInjectionSetup.cs
@@ -51,6 +51,7 @@ internal static class DependencyInjectionSetup
                 });
             })
             .AddHttpClient()
+            .AddLoggingHttpClients()
             .AddInitializers()
             .AddServiceCollection()
             .AddMediator(); // Add Mediator for event handling
--- a/src/ModularPipelines/Http/DurationLoggingHttpHandler.cs
+++ b/src/ModularPipelines/Http/DurationLoggingHttpHandler.cs
@@ -5,12 +5,12 @@ namespace ModularPipelines.Http;
 
 internal class DurationLoggingHttpHandler : DelegatingHandler
 {
-    private readonly IModuleLogger _logger;
+    private readonly IModuleLoggerProvider _loggerProvider;
     private readonly IHttpLogger _httpLogger;
 
-    public DurationLoggingHttpHandler(IModuleLogger logger, IHttpLogger httpLogger)
+    public DurationLoggingHttpHandler(IModuleLoggerProvider loggerProvider, IHttpLogger httpLogger)
     {
-        _logger = logger;
+        _loggerProvider = loggerProvider;
         _httpLogger = httpLogger;
     }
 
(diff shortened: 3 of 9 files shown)


thomhurst:
## Summary
Refactors HTTP logging to use IHttpClientFactory with pre-registered named clients instead of creating ServiceProviders per HttpLoggingType combination, fixing a memory leak issue.

## Critical Issues
None found ✅

## Suggestions

### 1. Handler lifecycle and logger context concern
**Location:** All logging handlers (RequestLoggingHttpHandler, ResponseLoggingHttpHandler, etc.)

The handlers are now registered as **transient** in `AddLoggingHttpClients()` (line 20-23), but they accept `IModuleLoggerProvider` in their constructors. Since handlers are created once per request and cached by IHttpClientFactory's handler pipeline, there's a subtle timing issue:

**Current flow:**
1. Handler is created (constructor runs) - gets IModuleLoggerProvider injected
2. Handler is reused for multiple requests
3. On each request, `GetLogger()` is called to get the current module's logger

**The issue:** This works correctly! The handlers properly defer logger resolution to request time by calling `_loggerProvider.GetLogger()` in `SendAsync`. This is the correct pattern for handlers that may be reused across different module contexts.

However, there's one edge case to verify: **What happens when HttpClients are used outside of a module execution context?** For example, if a service calls `IHttp.GetLoggingHttpClient()` during startup initialization or in a background task not tied to a module?

**Verification needed:** 
- Does `IModuleLoggerProvider.GetLogger()` handle being called outside module execution contexts gracefully (return a default logger or throw meaningful exception)?
- The existing tests in HttpTests.cs may not catch this since they use `GetService<IHttp>()` which likely sets up module context.

This isn't blocking since the original code had the same implicit dependency, but worth documenting or adding a guard.

### 2. Handler registration optimization
**Location:** HttpClientServiceCollectionExtensions.cs lines 19-23

All handlers are registered as transient globally, even though some logging type combinations don't use certain handlers (e.g., loggingType=0 uses no handlers). This creates unused service registrations.

**Impact:** Minimal - DI container overhead is negligible, but it's slightly inelegant.

**Alternative approach (optional):** Register handlers only once as transient before the loop, which you already do. Current implementation is fine.

### 3. Missing IDisposable removal in IHttp interface
**Location:** Http.cs

The PR removes the `IDisposable` implementation from the `Http` class, but I don't see a corresponding change removing `IDisposable` from the `IHttp` interface definition (if it was there).

**Verification needed:** Check if `IHttp` inherited from `IDisposable` - if so, the interface should also be updated.

## Verdict

The refactoring correctly addresses issue #1561 by eliminating per-combination ServiceProvider creation. The pattern of using IModuleLoggerProvider with lazy resolution is sound. The suggestions above are minor edge cases worth considering but don't block the PR.

PR #1765, opened 2026-01-02
refactor: Replace Task.Result with GetAwaiter().GetResult() (#1504)

### LINKED ISSUE #1504: Performance: ModuleRunner.ExecuteModuleLifecycle accesses Task.Result in catch block
## Issue Location
- **File**: `src/ModularPipelines/Engine/Execution/ModuleRunner.cs`
- **Lines**: 238-243

## Problem
In the catch block, the code accesses `.Result` on a task that has completed:

```csharp
catch (Exception ex)
{
    // Even when an exception is thrown, we need to register the result if one was set
    if (executionContext.ExecutionTask.IsCompleted && !executionContext.ExecutionTask.IsFaulted && !executionContext.ExecutionTask.IsCanceled)
    {
        var result = executionContext.ExecutionTask.Result;  // Blocking!
        moduleState.Result = result;
        _resultRegistry.RegisterResult(moduleType, result);
    }
    // ...
}
```

While `.Result` is safe here because we've verified `IsCompleted && !IsFaulted && !IsCanceled`, it:
1. Still uses the sync-over-async pattern which is an anti-pattern
2. Could be replaced with a non-blocking alternative
3. Sets a bad example in the codebase

## Suggested Fix
Use `GetAwaiter().GetResult()` or better yet, redesign to avoid this:

Option 1 - If we must access synchronously (e.g., in catch block):
```csharp
// This is marginally better as it doesn't wrap exceptions in AggregateException
var result = executionContext.ExecutionTask.GetAwaiter().GetResult();
```

Option 2 - Store the result proactively when it completes successfully:
The `ModuleExecutionContext` could cache the result when `SetResult` is called, avoiding the need to access the Task at all:

```csharp
// In ModuleExecutionContext
public IModuleResult? CachedResult { get; private set; }

public void SetResult(IModuleResult result)
{
    CachedResult = result;  // Cache for sync access
    _resultSource.TrySetResult(result);
}

// Then in catch block:
if (executionContext.CachedResult != null)
{
    moduleState.Result = executionContext.CachedResult;
    _resultRegistry.RegisterResult(moduleType, executionContext.CachedResult);
}
```

## Impact
Low - this only happens in exception scenarios and the task is already completed, but it's worth fixing to maintain clean async patterns throughout the codebase.

thomhurst:
## Summary
- Replaces `.Result` with `.GetAwaiter().GetResult()` in the catch block
- Avoids wrapping exceptions in AggregateException
- Follows better async patterns while maintaining sync access where required

The task is already verified to be completed and not faulted before access, so this is safe.

Fixes #1504

## Test plan
- [x] Code compiles successfully
- [ ] Existing tests pass (CI will verify)

### DIFF
--- a/src/ModularPipelines/Engine/Execution/ModuleRunner.cs
+++ b/src/ModularPipelines/Engine/Execution/ModuleRunner.cs
@@ -258,7 +258,8 @@ internal class ModuleRunner : IModuleRunner
             // Even when an exception is thrown, we need to register the result if one was set
             if (executionContext.ExecutionTask.IsCompleted && !executionContext.ExecutionTask.IsFaulted && !executionContext.ExecutionTask.IsCanceled)
             {
-                var result = executionContext.ExecutionTask.Result;
+                // Use GetAwaiter().GetResult() instead of .Result to avoid wrapping in AggregateException
+                var result = executionContext.ExecutionTask.GetAwaiter().GetResult();
                 moduleState.Result = result;
                 _resultRegistry.RegisterResult(moduleType, result);
             }


thomhurst:
## Summary
Replaces `Task.Result` with `GetAwaiter().GetResult()` in exception handling to avoid potential AggregateException wrapping.

## Critical Issues
None found ✅

## Suggestions

**Consider the more comprehensive fix from issue #1504**

The current change implements Option 1 (minimal fix), but Option 2 from the issue would be architecturally cleaner:

```csharp
// Cache result in ModuleExecutionContext when SetResult is called
public IModuleResult? CachedResult { get; private set; }

public void SetResult(IModuleResult result)
{
    CachedResult = result;
    _resultSource.TrySetResult(result);
}
```

Then in the catch block at src/ModularPipelines/Engine/Execution/ModuleRunner.cs:261:
```csharp
if (executionContext.CachedResult != null)
{
    moduleState.Result = executionContext.CachedResult;
    _resultRegistry.RegisterResult(moduleType, executionContext.CachedResult);
}
```

This would:
- Eliminate the Task access entirely
- Remove the need for `IsCompleted && !IsFaulted && !IsCanceled` checks
- Provide a cleaner sync access pattern
- Set a better example for the codebase

However, the current fix is valid and safe, so this is optional.

## Verdict

The change is correct and improves the code. The task state is verified before access, making this safe. While a more comprehensive architectural change could be cleaner (per issue #1504's Option 2), this minimal fix appropriately addresses the immediate concern.

Issue #1514, opened 2025-12-30
Performance: Command.BuildPlaceholderValueLookup uses reflection without caching


thomhurst:
## Issue Location
- **File**: `src/ModularPipelines/Context/Command.cs`
- **Lines**: 187-214

## Problem
The `BuildPlaceholderValueLookup` method uses reflection for every command execution:

```csharp
private static Dictionary<string, List<string>> BuildPlaceholderValueLookup(object optionsObject)
{
    var lookup = new Dictionary<string, List<string>>(StringComparer.Ordinal);
    var type = optionsObject.GetType();
    var properties = type.GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);

    foreach (var property in properties)
    {
        var cliArgument = property.GetCustomAttribute<CliArgumentAttribute>();
        if (cliArgument?.Name is null)
        {
            continue;
        }

        var rawValue = property.GetValue(optionsObject);  // Reflection call!
        // ...
    }
    return lookup;
}
```

Issues:
1. `GetProperties` called on every command execution
2. `GetCustomAttribute` called for every property on every execution
3. `GetValue` uses reflection for property access
4. No caching of placeholder property metadata

## Suggested Fix
Cache the placeholder property information per type:

```csharp
private static readonly ConcurrentDictionary<Type, PlaceholderPropertyInfo[]> PlaceholderCache = new();

private record PlaceholderPropertyInfo(string Name, Func<object, object?> Getter);

private static Dictionary<string, List<string>> BuildPlaceholderValueLookup(object optionsObject)
{
    var lookup = new Dictionary<string, List<string>>(StringComparer.Ordinal);
    var type = optionsObject.GetType();
    
    var placeholderProperties = PlaceholderCache.GetOrAdd(type, static t =>
    {
        return t.GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)
            .Select(p => (Property: p, Attr: p.GetCustomAttribute<CliArgumentAttribute>()))
            .Where(x => x.Attr?.Name != null)
            .Select(x => new PlaceholderPropertyInfo(
                x.Attr!.Name!,
                CreatePropertyGetter(x.Property)))
            .ToArray();
    });

    foreach (var prop in placeholderProperties)
    {
        var rawValue = prop.Getter(optionsObject);
        if (rawValue is null) continue;
        
        var values = GetArgumentValues(rawValue);
        if (values.Count > 0)
[...]

PR #1769, opened 2026-01-02
perf(command): Cache reflection in BuildPlaceholderValueLookup (#1514)

thomhurst:
## Summary
- Add `ConcurrentDictionary` cache for placeholder property metadata per options type
- Create compiled expression tree delegates for property access instead of using `PropertyInfo.GetValue`
- Replace per-call reflection (`GetProperties`, `GetCustomAttribute`, `GetValue`) with cached property getters

## Problem
The `BuildPlaceholderValueLookup` method in `Command.cs` uses reflection on every command execution:
- `GetProperties` called on every command execution
- `GetCustomAttribute` called for every property on every execution
- `GetValue` uses reflection for property access

This is inefficient for repeated command executions with the same options type.

## Solution
Cache the placeholder property information (name and compiled getter delegate) per type using a `ConcurrentDictionary`. The getter delegates are compiled from expression trees, providing near-native performance for property access.

The pattern is similar to how `CommandModelProvider` already caches its reflection data.

## Test plan
- [x] Build succeeds: `dotnet build src/ModularPipelines -c Release`
- [ ] Existing tests pass (run via CI)

### DIFF
--- a/src/ModularPipelines/Context/Command.cs
+++ b/src/ModularPipelines/Context/Command.cs
@@ -1,6 +1,8 @@
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
+using System.Linq.Expressions;
 using System.Reflection;
 using System.Text;
 using CliWrap;
@@ -16,6 +18,8 @@ namespace ModularPipelines.Context;
 
 public sealed class Command : ICommand
 {
+    private static readonly ConcurrentDictionary<Type, IReadOnlyList<PlaceholderPropertyInfo>> PlaceholderPropertyCache = new();
+
     private readonly ICommandLogger _commandLogger;
     private readonly ICommandModelProvider _commandModelProvider;
     private readonly ICommandArgumentBuilder _commandArgumentBuilder;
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Adds performance optimization by caching reflection metadata for placeholder property access using ConcurrentDictionary and compiled expression tree delegates.

## Critical Issues
None found ✅

## Suggestions

### Consider AOT Compatibility
The use of Expression.Compile() in CreatePropertyGetter (src/ModularPipelines/Context/Command.cs:246) may not be compatible with Native AOT compilation if that becomes a future goal. Expression tree compilation is not supported in AOT scenarios.

**Current impact**: None - the project does not currently target AOT.

**If AOT becomes a goal**, consider one of these alternatives:
- Use source generators to pre-generate property accessors at compile time
- Fall back to MethodInfo.CreateDelegate() for simple property getters (though this still has AOT limitations)
- Use [DynamicallyAccessedMembers] annotations and accept the reflection overhead

Since the project does not currently use AOT and this follows the same pattern as existing reflection-heavy code in the codebase, this is fine for now. Just flagging for awareness.

### Minor: XML doc completeness
The new PlaceholderPropertyInfo record (src/ModularPipelines/Context/Command.cs:386-387) has good documentation. Consider adding a remarks section explaining that getters are compiled expression trees for performance, which would help future maintainers understand the design choice.

## Verdict

The optimization is well-implemented and follows established patterns in the codebase. The caching strategy mirrors CommandModelProvider nicely, and the compiled expression trees should provide significant performance improvements for repeated command executions.

PR #1775, opened 2026-01-02
fix(filesystem): Add thread-safety and GC.SuppressFinalize to TempFile/TempFolder

thomhurst:
## Summary

- Use `Interlocked.Exchange` for thread-safe disposal check to prevent race conditions when both `Dispose()` and `DisposeAsync()` are called concurrently
- Add `GC.SuppressFinalize(this)` in both Dispose and DisposeAsync methods as per the standard IDisposable/IAsyncDisposable pattern

This addresses review feedback from PR #1771.

## Test plan

- [ ] Existing tests continue to pass
- [ ] Thread-safe disposal prevents double-delete race conditions

### DIFF
--- a/src/ModularPipelines/FileSystem/TempFile.cs
+++ b/src/ModularPipelines/FileSystem/TempFile.cs
@@ -22,7 +22,7 @@ namespace ModularPipelines.FileSystem;
 /// </remarks>
 public sealed class TempFile : IAsyncDisposable, IDisposable
 {
-    private bool _disposed;
+    private int _disposed;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="TempFile"/> class with a new temporary file path.
@@ -54,12 +54,10 @@ public sealed class TempFile : IAsyncDisposable, IDisposable
     /// <returns>A task that represents the asynchronous dispose operation.</returns>
     public async ValueTask DisposeAsync()
     {
-        if (_disposed)
+        if (Interlocked.Exchange(ref _disposed, 1) == 1)
         {
             return;
         }
-
-        _disposed = true;
 
         if (File.Exists)
         {
@@ -72,12 +70,10 @@ public sealed class TempFile : IAsyncDisposable, IDisposable
     /// </summary>
     public void Dispose()
     {
-        if (_disposed)
+        if (Interlocked.Exchange(ref _disposed, 1) == 1)
         {
             return;
         }
-
-        _disposed = true;
 
         if (File.Exists)
         {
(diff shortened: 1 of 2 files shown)


thomhurst:
## Summary
This PR makes the dispose pattern in `TempFile` and `TempFolder` thread-safe by replacing the boolean flag with an interlocked integer exchange.

## Critical Issues

**Race condition between Dispose() and DisposeAsync()**

While the PR correctly makes the dispose check thread-safe, there's still a race condition when both `Dispose()` and `DisposeAsync()` are called concurrently (or when either is called from multiple threads).

The current implementation in both files:
```csharp
public async ValueTask DisposeAsync()
{
    if (Interlocked.Exchange(ref _disposed, 1) == 1)
    {
        return;
    }

    if (File.Exists)  // or Folder.Exists
    {
        await File.DeleteAsync().ConfigureAwait(false);
    }
}

public void Dispose()
{
    if (Interlocked.Exchange(ref _disposed, 1) == 1)
    {
        return;
    }

    if (File.Exists)  // or Folder.Exists
    {
        File.Delete();
    }
}
```

**Problem**: If thread A calls `DisposeAsync()` and thread B calls `Dispose()` at nearly the same time:
1. Thread A: `Interlocked.Exchange` returns 0, proceeds
2. Thread B: `Interlocked.Exchange` returns 1, exits ✓
3. Thread A: Checks `File.Exists` (true), starts async delete
4. BUT if Thread B won the race: Thread B checks `File.Exists` (true), starts sync delete
5. Both deletion operations execute concurrently → potential exceptions or file system errors

**Solution**: The `.NET dispose pattern` for types implementing both `IDisposable` and `IAsyncDisposable` typically has `Dispose()` call `DisposeAsync().GetAwaiter().GetResult()` or use a synchronous-only code path with proper locking.

Example fix:
```csharp
public void Dispose()
{
    if (Interlocked.Exchange(ref _disposed, 1) == 1)
    {
        return;
    }

    if (File.Exists)
    {
        File.Delete();
    }

    GC.SuppressFinalize(this);
}

public async ValueTask DisposeAsync()
{
    if (Interlocked.Exchange(ref _disposed, 1) == 1)
    {
        return;
    }

    if (File.Exists)
    {
        await File.DeleteAsync().ConfigureAwait(false);
    }

    GC.SuppressFinalize(this);
}
```

This still has the same issue. A better approach would be to use a lock or ensure mutual exclusion between the two paths. However, since these are temp files typically used in `using` statements, the real question is: **Is concurrent disposal a realistic scenario for these classes?**

If not, the current PR is acceptable. If yes, additional synchronization is needed.

Also note: Both methods should probably call `GC.SuppressFinalize(this)` even though there's no finalizer currently.

## Suggestions
None - the thread-safety improvement is good, assuming concurrent Dispose/DisposeAsync isn't a concern.

## Verdict
⚠️ **REQUEST CHANGES** - Need clarification on the Dispose/DisposeAsync race condition scenario, or additional synchronization if concurrent disposal is possible.

PR #1780, opened 2026-01-02
feat(http): Add retry policy for transient HTTP failures

### LINKED ISSUE #1516: Code smell: No HTTP retry policy configured for transient failures
## Issue

**Location:** 
- `C:\git\ModularPipelines\src\ModularPipelines\DependencyInjection\DependencyInjectionSetup.cs` (line 53)
- `C:\git\ModularPipelines\src\ModularPipelines\Http\Http.cs` (lines 52-103)

**Problem:**
The HTTP client setup uses `.AddHttpClient()` without configuring any retry policies for transient HTTP failures. While the framework has Polly as a dependency and uses it for module-level retries, the HTTP client itself does not benefit from transient fault handling.

**Current code (DependencyInjectionSetup.cs line 53):**
```csharp
.AddHttpClient()  // No retry policy configured
```

**What's missing:**
- No retry for transient HTTP errors (408, 429, 500, 502, 503, 504)
- No retry for network failures (HttpRequestException)
- No exponential backoff between retries
- No circuit breaker pattern for failing services

**Risk:** MEDIUM - HTTP calls to external services will fail immediately on transient issues instead of retrying, causing unnecessary pipeline failures.

**Suggested Fix:**
Add a transient HTTP error policy using Microsoft.Extensions.Http.Polly:

```csharp
// In DependencyInjectionSetup.cs
services.AddHttpClient()
    .AddTransientHttpErrorPolicy(policy => 
        policy.WaitAndRetryAsync(3, retryAttempt => 
            TimeSpan.FromSeconds(Math.Pow(2, retryAttempt))))
    .AddTransientHttpErrorPolicy(policy => 
        policy.CircuitBreakerAsync(5, TimeSpan.FromSeconds(30)));
```

**Alternative:** Make retry configurable via `HttpOptions` so users can opt-in/out per request:
```csharp
public record HttpOptions(HttpRequestMessage HttpRequestMessage)
{
    public int RetryCount { get; init; } = 0;  // 0 = no retry, or use pipeline default
    public TimeSpan RetryDelay { get; init; } = TimeSpan.FromSeconds(1);
}
```

**Note:** The framework already has `PipelineOptions.DefaultRetryCount` for modules - a similar pattern could be applied to HTTP requests.

thomhurst:
## Summary
- Adds `HttpResilienceOptions` record with configurable retry settings
- Creates `ResilienceHttpHandler` using Polly for retry with exponential backoff
- Integrates retry handler into HTTP client via DI registration
- Provides preset configurations: Default, None, Aggressive, Conservative

Closes #1516

## Test plan
- [ ] Verify transient failures (5xx, 408, 429) trigger retries
- [ ] Verify non-retryable status codes (4xx) don't retry
- [ ] Verify exponential backoff and jitter work correctly
- [ ] Run existing tests to confirm no regressions

### DIFF
--- a/src/ModularPipelines/Http/HttpClientServiceCollectionExtensions.cs
+++ b/src/ModularPipelines/Http/HttpClientServiceCollectionExtensions.cs
@@ -3,7 +3,7 @@ using Microsoft.Extensions.DependencyInjection;
 namespace ModularPipelines.Http;
 
 /// <summary>
-/// Extension methods for registering logging HttpClients with the DI container.
+/// Extension methods for registering logging and resilience HttpClients with the DI container.
 /// </summary>
 internal static class HttpClientServiceCollectionExtensions
 {
@@ -11,6 +11,7 @@ internal static class HttpClientServiceCollectionExtensions
     /// Registers all 16 combinations of logging HttpClients (2^4 from the 4 HttpLoggingType flags).
     /// This allows sharing HttpClient instances through IHttpClientFactory instead of creating
     /// new ServiceProviders per logging type configuration.
+    /// Also adds resilience handlers for retry policies on transient failures.
     /// </summary>
     /// <param name="services">The service collection.</param>
     /// <returns>The service collection for chaining.</returns>
@@ -21,6 +22,7 @@ internal static class HttpClientServiceCollectionExtensions
         services.AddTransient<ResponseLoggingHttpHandler>();
         services.AddTransient<DurationLoggingHttpHandler>();
         services.AddTransient<StatusCodeLoggingHttpHandler>();
+        services.AddTransient<ResilienceHttpHandler>();
 
         // Register all 16 combinations of logging types (0-15)
         // HttpLoggingType is a flags enum with: Request=1, Response=2, StatusCode=4, Duration=8
@@ -30,6 +32,10 @@ internal static class HttpClientServiceCollectionExtensions
             var clientName = HttpClientNames.GetClientName(loggingType);
 
             var builder = services.AddHttpClient(clientName);
+
+            // Add resilience handler first (outermost) so retries wrap all other handlers
+            // This ensures logging happens for each attempt
+            builder.AddHttpMessageHandler<ResilienceHttpHandler>();
 
             // Add handlers based on the flags
             if (loggingType.HasFlag(HttpLoggingType.Request))
--- a/src/ModularPipelines/Http/ResilienceHttpHandler.cs
+++ b/src/ModularPipelines/Http/ResilienceHttpHandler.cs
@@ -1,0 +1,170 @@
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using ModularPipelines.Logging;
+using ModularPipelines.Options;
+using Polly;
+using Polly.Retry;
+
+namespace ModularPipelines.Http;
+
+/// <summary>
+/// A delegating handler that adds retry resilience to HTTP requests using Polly.
+/// Handles transient failures including network errors and server errors (5xx).
+/// </summary>
+internal class ResilienceHttpHandler : DelegatingHandler
+{
+    private readonly IModuleLoggerProvider _loggerProvider;
+    private readonly IOptions<PipelineOptions> _pipelineOptions;
+
+    public ResilienceHttpHandler(
+        IModuleLoggerProvider loggerProvider,
+        IOptions<PipelineOptions> pipelineOptions)
+    {
+        _loggerProvider = loggerProvider;
+        _pipelineOptions = pipelineOptions;
+    }
+
+    /// <inheritdoc/>
+    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        var options = _pipelineOptions.Value.DefaultHttpResilienceOptions ?? HttpResilienceOptions.Default;
+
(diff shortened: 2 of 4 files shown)


thomhurst:
## Summary
Adds HTTP retry resilience using Polly with configurable retry policies for transient failures (network errors, 5xx server errors).

## Critical Issues

### 1. Thread-Safety: Random Instance Not Thread-Safe ❌
**File:** `src/ModularPipelines/Http/ResilienceHttpHandler.cs:18`

The `_random` field is shared across all requests but `Random` is not thread-safe. When multiple HTTP requests execute concurrently, this can cause race conditions.

**Fix:** Use `Random.Shared` (available in .NET 6+) or `ThreadLocal<Random>`:

```csharp
// Option 1: For .NET 6+ (preferred)
private TimeSpan CalculateDelay(int retryAttempt, HttpResilienceOptions options)
{
    // ... existing code ...
    if (options.JitterFactor > 0)
    {
        var jitter = baseDelay.TotalMilliseconds * options.JitterFactor * ((Random.Shared.NextDouble() * 2) - 1);
        // ...
    }
}

// Option 2: For multi-targeting frameworks
private static readonly ThreadLocal<Random> _random = new(() => new Random());

// Then use: _random.Value.NextDouble()
```

---

### 2. HttpContent Cannot Be Reused Across Retries ❌
**File:** `src/ModularPipelines/Http/ResilienceHttpHandler.cs:130`

The `CloneRequest` method assigns the original content directly:
```csharp
if (original.Content != null)
{
    clone.Content = original.Content;  // ❌ Content already consumed
}
```

**Problem:** After the first request attempt, the content stream is consumed and cannot be read again. Retries will fail with "Cannot access a disposed object" or "Stream was not readable".

**Fix:** The comment acknowledges this ("Streaming content cannot be retried") but the code doesn't handle it. Options:
1. Don't clone content at all and document that retries only work with non-streaming content
2. Add a warning when content exists
3. Try to buffer content if possible

**Suggested fix:**
```csharp
// Note: HttpContent can only be sent once. For retries to work with request bodies,
// the content must be bufferable (not a stream that's already been consumed).
// This is a known limitation - streaming content cannot be retried.
if (original.Content != null)
{
    // Only assign content reference - caller must ensure content is reusable
    // For streaming scenarios, content will be consumed on first attempt
    clone.Content = original.Content;
}
```

Or better yet, document this limitation in the XML docs for `HttpResilienceOptions` so users know retries won't work with streaming request bodies.

---

## Suggestions

### 1. Consider Polly v8+ Resilience Pipelines
Polly v8 introduced a new resilience pipeline API that's more modern and efficient. The current implementation uses the older v7 API. This is fine but consider migrating to the newer API if you bump to Polly v8+.

### 2. Document Content Reuse Limitation
Add to `HttpResilienceOptions` XML docs:
```csharp
/// <remarks>
/// <para>⚠️ Limitation: Retries do not work with streaming request bodies (POST/PUT with streams).
/// HttpContent can only be consumed once. For retry support, use buffered content or small payloads.</para>
/// </remarks>
```

---

## Previous Review Status
No previous comments found.

---

## Verdict
⚠️ **REQUEST CHANGES** - Critical thread-safety and content reuse issues must be fixed before merging.

thomhurst:
## Summary
Adds HTTP resilience support with configurable retry policies for transient HTTP failures using Polly.

## Critical Issues
None found ✅

## Suggestions

### 1. Request Content Buffering - Performance Consideration
**Location:** src/ModularPipelines/Http/ResilienceHttpHandler.cs:41-43

The implementation buffers all request content into memory unconditionally before any retries:
```csharp
if (request.Content != null)
{
    contentBytes = await request.Content.ReadAsByteArrayAsync(cancellationToken).ConfigureAwait(false);
    contentType = request.Content.Headers.ContentType?.ToString();
}
```

**Concern:** This could cause significant memory pressure for large file uploads or streaming scenarios, even when no retry is actually needed. The content is buffered before the first attempt, not just on retries.

**Suggestion:** Consider documenting this limitation more prominently in the XML docs (it's mentioned in HttpResilienceOptions remarks, but not on ResilienceHttpHandler itself). Alternatively, consider lazy buffering - only buffer after the first failure, and fail subsequent retries if the content wasn't seekable.

### 2. Handler Ordering in Documentation
**Location:** src/ModularPipelines/Http/HttpClientServiceCollectionExtensions.cs:36-37

Good comment about resilience handler being first (outermost), but this creates a subtle ordering where:
- Resilience wraps everything (each retry logs separately) ✅
- But duration logging will measure total time across all retries, not per-attempt duration

This is probably intentional, but worth confirming the desired behavior. If you want per-attempt duration logging, the handlers would need reordering.

### 3. Minor: Polly Context Unused
**Location:** src/ModularPipelines/Http/ResilienceHttpHandler.cs:98

The `Polly.Context` parameter in `OnRetry` is unused. While not harmful, you could remove it or use it to pass request-specific metadata for debugging.

### 4. Consider: Idempotency Safety
The current implementation will retry all HTTP methods including non-idempotent ones (POST, PATCH, etc.). While the PR doesn't claim to handle this, consider whether you want to add an option to restrict retries to idempotent methods (GET, PUT, DELETE, HEAD, OPTIONS) by default.

Example:
```csharp
public bool OnlyRetryIdempotentRequests { get; init; } = true; // Safe default
```

This would prevent accidentally retrying a payment twice, for example.

## Verdict

The implementation is solid and follows existing patterns in the codebase. The suggestions above are optional improvements focused on documentation clarity and potential edge cases, not blocking issues.

Issue #1554, opened 2025-12-30
Code smell: GitHub client manually chains HttpMessageHandlers instead of using IHttpClientFactory


thomhurst:
## Issue

**Location:** C:\git\ModularPipelines\src\ModularPipelines.GitHub\GitHub.cs (lines 71-86)

**Problem:**
The GitHub client creates HttpMessageHandler chains manually instead of using the IHttpClientFactory pattern. This creates several issues:

1. HttpClientHandler is created directly without pooling
2. Handler chain is manually constructed which is error-prone
3. Not benefiting from IHttpClientFactory connection management
4. Potential socket exhaustion under high load

**Current code (GitHub.cs lines 71-86):**
var connection = new Connection(new ProductHeaderValue("ModularPipelines"),
  new HttpClientAdapter(() =>
  {
      var moduleLogger = _moduleLoggerProvider.GetLogger();
      return new RequestLoggingHttpHandler(moduleLogger, _httpLogger)
      {
          InnerHandler = new ResponseLoggingHttpHandler(moduleLogger, _httpLogger)
          {
              InnerHandler = new StatusCodeLoggingHttpHandler(moduleLogger, _httpLogger)
              {
                  InnerHandler = new HttpClientHandler(),  // Direct instantiation
              },
          },
      };
  }));

**Risk:** MEDIUM - Socket exhaustion possible under heavy load, manual handler management is fragile.

**Suggested Fix:**
1. Create a named HttpClient registration using IHttpClientFactory
2. Configure message handlers via AddHttpMessageHandler
3. Inject the configured HttpClient into the GitHub client

Example:
services.AddHttpClient("GitHub")
    .AddHttpMessageHandler<RequestLoggingHttpHandler>()
    .AddHttpMessageHandler<ResponseLoggingHttpHandler>()
    .AddHttpMessageHandler<StatusCodeLoggingHttpHandler>();

Then inject IHttpClientFactory and use CreateClient("GitHub") instead of manual construction.

PR #1784, opened 2026-01-02
refactor(git): Use SocketsHttpHandler with connection pooling for GitHub client (#1554)

thomhurst:
## Summary
Refactors GitHub client HTTP handler creation to use `SocketsHttpHandler` with connection pooling to prevent socket exhaustion.

## Changes
- Switch from `HttpClientHandler` to `SocketsHttpHandler` with connection pooling configuration
- Extract handler chain creation into a dedicated `CreateHandler()` method
- Configure `PooledConnectionLifetime` (2 min), `PooledConnectionIdleTimeout` (1 min), and `MaxConnectionsPerServer` (20)
- Each `HttpClientAdapter` gets its own handler chain, avoiding thread-safety issues with shared `DelegatingHandler.InnerHandler` references

## Why Not IHttpClientFactory?
Octokit's `HttpClientAdapter` requires a `Func<HttpMessageHandler>`, not an `HttpClient`. The standard `IHttpClientFactory` pattern returns `HttpClient` instances, which don't directly integrate with Octokit's `Connection` class.

The approach used here:
1. Creates fresh handler chains per invocation (avoiding shared `DelegatingHandler` issues)
2. Uses `SocketsHttpHandler` which provides connection pooling internally via `PooledConnectionLifetime`
3. Keeps the implementation simple without requiring custom factories or complex DI wiring

## Test plan
- [ ] Verify GitHub client still works correctly
- [ ] Verify logging handlers are applied


thomhurst:
## Summary
Changes GitHub client from HttpClientHandler to SocketsHttpHandler with connection pooling settings and extracts handler creation into a separate method.

## Critical Issues

### 1. Misleading PR Title and Description ❌
The PR title claims "Use IHttpClientFactory for GitHub client" but the implementation **does not use IHttpClientFactory at all**. The changes only:
- Switch from `HttpClientHandler` to `SocketsHttpHandler`
- Extract handler creation into `CreateHandler()` method
- Add connection pooling configuration

Issue #1554 specifically suggested using `IHttpClientFactory` with `AddHttpClient` and `AddHttpMessageHandler`, which this PR does not implement.

**Location:** PR title, body, and src/ModularPipelines.GitHub/GitHub.cs:67-106

### 2. Doesn't Actually Fix Issue #1554 ❌
The referenced issue requested IHttpClientFactory integration, not just switching to SocketsHttpHandler. While SocketsHttpHandler does provide connection pooling, this doesn't address the core ask for proper dependency injection and factory pattern usage.

**Location:** Issue #1554 vs actual implementation

### 3. Potential Thread Safety Concerns ⚠️
The `CreateHandler()` method is called via lambda from `HttpClientAdapter`. Each call creates new instances of the DelegatingHandler chain. If Octokit's `HttpClientAdapter` calls this factory from multiple threads, there could be issues with:
- Multiple handler instances being created unnecessarily
- Potential race conditions if handlers share state

The comment "we don't need to cache handler instances" needs verification - does Octokit's `Connection` class ensure thread-safe, single-call initialization of the adapter?

**Location:** src/ModularPipelines.GitHub/GitHub.cs:90 (CreateHandler method)

### 4. Removed Using Statement
The PR removes `using Microsoft.Extensions.Http.Logging;` but doesn't explain why. Was this unused?

**Location:** src/ModularPipelines.GitHub/GitHub.cs:1

## Suggestions

1. **If you want to use IHttpClientFactory properly:**
   - Register a named HttpClient in `GitHubExtensions.cs`
   - Use `AddHttpMessageHandler` for the logging handlers
   - Inject `IHttpClientFactory` into the `GitHub` class
   - Use the factory to get an HttpClient instance
   - Adapt the HttpClient for Octokit's Connection

2. **If SocketsHttpHandler is the intended approach:**
   - Update PR title/description to reflect what it actually does
   - Close #1554 and create a new issue explaining why IHttpClientFactory isn't suitable
   - Verify Octokit's `HttpClientAdapter` thread-safety behavior
   - Consider caching the handler instance if CreateHandler is called multiple times

3. **Connection pooling values need justification:**
   - Why 2-minute lifetime and 1-minute idle timeout?
   - Why max 20 connections per server?
   - Should these be configurable via GitHubOptions?

## Verdict
⚠️ **REQUEST CHANGES** - PR does not implement what it claims (IHttpClientFactory) and may not solve issue #1554 as intended

PR #1790, opened 2026-01-02
refactor(amazon): Encapsulate S3 client with wrapper methods (#1636)

### LINKED ISSUE #1636: Code smell: AWS S3 client exposed publicly without encapsulation
## Issue

**Location:** C:\git\ModularPipelines\src\ModularPipelines.AmazonWebServices\S3.cs (lines 10-15)

**Problem:**
The S3 class exposes the underlying AmazonS3Client directly through a public property. This violates encapsulation and allows callers to bypass any wrapper methods, logging, or future enhancements.

**Current code:**
public class S3
{
    public AmazonS3Client Client { get; }  // Publicly exposed!

    public S3(AmazonS3Client client)
    {
        Client = client;
    }
    // ... wrapper methods
}

**Issues:**
1. Callers can use Client directly, bypassing wrapper methods
2. Cannot add cross-cutting concerns (logging, metrics, validation) to all S3 operations
3. No abstraction - tied directly to AWS SDK types
4. Difficult to mock for testing

**Risk:** LOW - This is a design smell rather than a bug.

**Suggested Fix:**
Option 1: Make the client internal or private:
internal AmazonS3Client Client { get; }

Option 2: Extract an interface and hide the implementation:
public interface IS3
{
    Task<PutBucketResponse> Bucket(PutBucketRequest request, CancellationToken token = default);
    // ... other methods
}

Option 3: If public access is intentional, document it clearly:
/// <summary>
/// The underlying AWS S3 client. Use for operations not yet wrapped by this class.
/// Note: Direct usage bypasses any framework-level logging or error handling.
/// </summary>
public AmazonS3Client Client { get; }

thomhurst:
## Summary
- Made underlying AmazonS3Client internal instead of publicly exposed
- Added proper wrapper methods for common S3 operations
- Enables future cross-cutting concerns (logging, metrics, validation)
- Better testability through abstraction

Fixes #1636

## Test plan
- [ ] Build succeeds
- [ ] S3 operations work through wrapper methods
- [ ] Existing tests pass

### DIFF
--- a/src/ModularPipelines.AmazonWebServices/IS3.cs
+++ b/src/ModularPipelines.AmazonWebServices/IS3.cs
@@ -10,7 +10,7 @@ namespace ModularPipelines.AmazonWebServices;
 /// for common S3 operations including bucket management, object operations, and
 /// configuration settings.
 /// </remarks>
-public interface IS3
+public interface IS3 : IDisposable
 {
     #region Bucket Operations
 
--- a/src/ModularPipelines.AmazonWebServices/S3.cs
+++ b/src/ModularPipelines.AmazonWebServices/S3.cs
@@ -13,9 +13,10 @@ namespace ModularPipelines.AmazonWebServices;
 /// configuration settings.
 /// </remarks>
 [ExcludeFromCodeCoverage]
-public class S3 : IS3
+public class S3 : IS3, IDisposable
 {
-    private AmazonS3Client _client;
+    private readonly AmazonS3Client _client;
+    private bool _disposed;
 
     /// <summary>
     /// Gets the underlying Amazon S3 client.
@@ -522,6 +523,38 @@ public class S3 : IS3
     public async Task<PutBucketIntelligentTieringConfigurationResponse> BucketIntelligentTieringConfiguration(PutBucketIntelligentTieringConfigurationRequest request, CancellationToken cancellationToken = default)
     {
         return await PutBucketIntelligentTieringConfigurationAsync(request, cancellationToken).ConfigureAwait(false);
+    }
+
+    #endregion
+
+    #region IDisposable
+
+    /// <summary>
+    /// Disposes the underlying Amazon S3 client.
+    /// </summary>
+    public void Dispose()
+    {
+        Dispose(disposing: true);
+        GC.SuppressFinalize(this);
+    }
+
+    /// <summary>
+    /// Disposes the underlying Amazon S3 client.
+    /// </summary>
+    /// <param name="disposing">True if called from Dispose(), false if called from finalizer.</param>
+    protected virtual void Dispose(bool disposing)
+    {
+        if (_disposed)
+        {
+            return;
+        }
+
+        if (disposing)
+        {
+            _client.Dispose();
+        }
+
+        _disposed = true;
     }
 
     #endregion


thomhurst:
## Summary
This PR refactors the S3 class by introducing an IS3 interface, making the underlying AmazonS3Client private, renaming methods to follow async naming conventions, adding comprehensive XML documentation, and providing backward compatibility through `[Obsolete]` attributes.

## Critical Issues

### 1. Breaking Change: Client Property Removed (Not Addressed)
**Location:** src/ModularPipelines.AmazonWebServices/S3.cs:10

The `Client` property changed from `public` to `private` with **no backward compatibility shim**. While old methods have `[Obsolete]` attributes, the Client property does not.

**Impact:** Any external code accessing `s3.Client` directly will break immediately without migration path.

**Recommendation:** Add a deprecated public property:
```csharp
[Obsolete("Direct access to the underlying client is deprecated. Use the wrapper methods instead. This property will be removed in a future version.")]
public AmazonS3Client Client => _client;
private readonly AmazonS3Client _client;
```

This gives consumers a migration path and warning, rather than immediate breakage.

---

## Suggestions

### 1. Inconsistent Property Naming Pattern
**Location:** src/ModularPipelines.AmazonWebServices/S3.cs:10

Using `Client` as an auto-property and `_client` as backing field would require renaming the current usage. Consider simpler approach: keep current private property name but add public obsolete wrapper (see above).

### 2. Exception Handling Documentation
**Locations:** 
- src/ModularPipelines.AmazonWebServices/S3.cs (DoesBucketExistAsync)
- src/ModularPipelines.AmazonWebServices/S3.cs (DoesObjectExistAsync)

The methods catch only `StatusCode.NotFound` and return false. Other exceptions (403 Forbidden, network errors) will throw. This is likely correct behavior, but consider adding XML doc `<exception>` tags to document what exceptions consumers should expect:

```csharp
/// <exception cref="AmazonS3Exception">Thrown when access is denied or other S3 errors occur (not including NotFound).</exception>
```

---

## Previous Review Status

Both previous review comments raised the breaking change concern. This PR **partially addresses** it:
- ✅ **Fixed:** Old method names have `[Obsolete]` attributes for graceful migration
- ❌ **Not Fixed:** Client property removal is still a hard breaking change with no migration path

---

## Verdict
⚠️ **REQUEST CHANGES** - The Client property breaking change needs a backward compatibility shim.

**Reasoning:** While the method deprecations are well-handled, removing public API access without an `[Obsolete]` transition period is a hard breaking change for downstream consumers. Adding a deprecated public Client property would maintain backward compatibility while guiding users to the new wrapper methods.

thomhurst:
## Summary
This PR encapsulates the S3 client by making it internal and adding an `IS3` interface with wrapper methods for common S3 operations.

## Critical Issues

**1. Breaking change without proper versioning consideration** ⚠️  
The `Client` property is marked as `[Obsolete]` rather than being removed, which is correct for a transitional approach. However, consumers currently using `s3.Client.SomeMethod()` will need migration guidance. The obsolete message should include specific migration instructions.

**Location:** `src/ModularPipelines.AmazonWebServices/S3.cs:28`

**2. Potential null reference issue with private field** ⚠️  
The `_client` field is declared without nullable annotation and without `readonly`, but the obsolete `Client` property getter does not handle potential null cases. While the constructor initializes it, the mutable field could theoretically be set to null if there were any internal assignment logic.

**Location:** `src/ModularPipelines.AmazonWebServices/S3.cs:17`  
**Recommendation:** Make `_client` readonly: `private readonly AmazonS3Client _client;`

**3. Missing disposal pattern** ⚠️  
The `AmazonS3Client` implements `IDisposable`, but the `S3` wrapper class does not. Since `S3` is registered as scoped in DI (`services.TryAddScoped<IS3, S3>()`), and it wraps a disposable resource, it should implement `IDisposable` to properly dispose of the underlying client.

**Locations:**
- `src/ModularPipelines.AmazonWebServices/S3.cs:13` (class definition)
- `src/ModularPipelines.AmazonWebServices/Extensions/AmazonExtensions.cs:27` (DI registration)

## Suggestions

1. **Improve obsolete message with migration example** - Add specific examples to the obsolete attribute message to help consumers migrate.

2. **Consider adding XML docs example usage to IS3 interface** - The interface has good documentation, but adding example tags to a few key methods would help consumers understand the intended usage pattern.

3. **ConfigureAwait(false) is consistently used** ✅ Good practice throughout the implementation.

## Previous Review Status
Cannot access review comments due to GitHub token scope limitations, but based on the PR description this addresses issue #1636.

## Verdict
⚠️ **REQUEST CHANGES** - Critical issue #3 (missing disposal pattern) must be fixed to prevent resource leaks. Issue #2 (readonly field) is a defensive improvement that should be addressed.

PR #1793, opened 2026-01-02
refactor(engine): Improve context scope cleanup with ConcurrentBag and IAsyncDisposable (#1477)

### LINKED ISSUE #1477: Code smell: ModuleContextProvider accumulates scopes without cleanup
## Location
- **File**: `src/ModularPipelines/Engine/ModuleContextProvider.cs`
- **Lines**: 10, 17-23

## Problem
The `ModuleContextProvider` class maintains a list of service scopes that are never disposed:

```csharp
private readonly List<IServiceScope> _scopes = new();

public IPipelineContext GetModuleContext()
{
    var serviceScope = _serviceProvider.CreateAsyncScope();
    _scopes.Add(serviceScope);
    return serviceScope.ServiceProvider.GetRequiredService<IPipelineContext>();
}
```

## Why it's a problem
1. **Potential Resource Leak**: Service scopes are added to the list but there's no clear disposal mechanism
2. **Memory Growth**: The list grows unbounded during pipeline execution
3. **IScopeDisposer Interface**: While the class implements `IScopeDisposer`, the `GetScopes()` method only returns the scopes - it doesn't dispose them
4. **Thread Safety**: The List<T> is not thread-safe, but `GetModuleContext()` could be called from multiple threads

## Suggested Fix
1. Make the class implement `IDisposable` or `IAsyncDisposable`
2. Dispose all scopes when the provider is disposed
3. Use a thread-safe collection like `ConcurrentBag<IServiceScope>`

```csharp
internal class ModuleContextProvider : IPipelineContextProvider, IScopeDisposer, IAsyncDisposable
{
    private readonly IServiceProvider _serviceProvider;
    private readonly ConcurrentBag<IServiceScope> _scopes = new();

    // ...

    public async ValueTask DisposeAsync()
    {
        foreach (var scope in _scopes)
        {
            if (scope is IAsyncDisposable asyncDisposable)
                await asyncDisposable.DisposeAsync();
            else
                scope.Dispose();
        }
    }
}
```

## Category
- [ ] Magic numbers/strings
- [ ] DRY violation
- [ ] SOLID violation
- [x] Resource leaks
- [x] Thread safety issues

thomhurst:
## Summary
- Change `List<IServiceScope>` to `ConcurrentBag<IServiceScope>` for thread safety
- Implement `IAsyncDisposable` with `DisposeAsync()` for proper async cleanup
- Handle both synchronous and asynchronous disposal patterns

Fixes #1477

## Test Plan
- [x] Build succeeds
- [ ] Unit tests pass

### DIFF
--- a/src/ModularPipelines/Engine/ModuleContextProvider.cs
+++ b/src/ModularPipelines/Engine/ModuleContextProvider.cs
@@ -19,11 +19,18 @@ internal class ModuleContextProvider : IPipelineContextProvider, IScopeDisposer,
 
     public IPipelineContext GetModuleContext()
     {
-        ObjectDisposedException.ThrowIf(_disposed, this);
+        AsyncServiceScope serviceScope;
 
-        var serviceScope = _serviceProvider.CreateAsyncScope();
+        lock (_disposeLock)
+        {
+            ObjectDisposedException.ThrowIf(_disposed, this);
 
-        _scopes.Add(serviceScope);
+            // Create and add scope inside lock to ensure atomicity with disposal.
+            // This prevents a race where disposal could take a snapshot between
+            // the _disposed check and the Add operation, leaving the scope undisposed.
+            serviceScope = _serviceProvider.CreateAsyncScope();
+            _scopes.Add(serviceScope);
+        }
 
         return serviceScope.ServiceProvider.GetRequiredService<IPipelineContext>();
     }


thomhurst:
## Summary
This PR improves thread safety and resource cleanup in ModuleContextProvider by using ConcurrentBag and implementing proper async disposal.

## Critical Issues

### 1. Race condition between GetModuleContext and disposal
**Location**: ModuleContextProvider.cs:19-24, 33-60

There's a race condition window between checking `_disposed` and taking a snapshot of scopes:

```csharp
// Thread A (DisposeAsync)
lock (_disposeLock) {
    _disposed = true;  // Sets flag
}
// Lock released - Thread B can now enter GetModuleContext

// Thread B (GetModuleContext) 
ObjectDisposedException.ThrowIf(_disposed, this);  // Throws

// Thread A continues
var scopesToDispose = _scopes.ToArray();  // Takes snapshot
```

However, there's a subtle issue: if Thread B passes the check just before Thread A sets `_disposed = true`, it will add to `_scopes` after the snapshot is taken in DisposeAsync. This newly added scope would not be disposed.

**Fix**: Move the snapshot inside the lock to ensure atomicity:

```csharp
public async ValueTask DisposeAsync()
{
    IServiceScope[] scopesToDispose;
    
    lock (_disposeLock)
    {
        if (_disposed)
        {
            return;
        }
        _disposed = true;
        scopesToDispose = _scopes.ToArray();  // Snapshot inside lock
    }

    foreach (var scope in scopesToDispose)
    {
        await ((IAsyncDisposable)scope).DisposeAsync().ConfigureAwait(false);
    }

    GC.SuppressFinalize(this);
}
```

Apply the same pattern to the synchronous `Dispose()` method.

### 2. Potential ObjectDisposedException leak in GetScopes()
**Location**: ModuleContextProvider.cs:30-38

The `GetScopes()` method returns an empty array when disposed to prevent double-disposal. However, this silently swallows the disposed state. Since `IScopeDisposer.Dispose()` (defined in the interface with a default implementation) calls `GetScopes()`, it won't properly dispose scopes if called.

**Impact**: If someone calls the interface's default `Dispose()` method instead of the explicit implementation, scopes won't be disposed.

**Recommendation**: Either:
1. Remove the early return and let it throw, OR
2. Document why returning empty is safe (because the explicit `IDisposable.Dispose` is what gets called)

Given the explicit `IDisposable.Dispose` implementation, option 2 seems intended, but needs a comment explaining this.

## Suggestions

None - the critical issues above must be addressed first.

## Verdict
⚠️ **REQUEST CHANGES** - Critical race condition and disposal semantics issues found

thomhurst:
## Summary
Adds thread-safe scope management with ConcurrentBag and implements IAsyncDisposable for proper async cleanup of service scopes.

## Critical Issues

### 1. Race condition between GetModuleContext() and disposal (src/ModularPipelines/Engine/ModuleContextProvider.cs:21)
There's a race condition where a thread could pass the _disposed check in GetModuleContext() (line 22), then get suspended before adding to _scopes (line 26). Meanwhile, another thread enters disposal, takes the snapshot, and completes. The first thread then resumes and adds a scope that will never be disposed.

Suggested fix: Move the _disposed check inside the lock to ensure atomicity.

### 2. Double disposal via IScopeDisposer.Dispose() (src/ModularPipelines/Interfaces/IScopeDisposer.cs:19-25)
The IScopeDisposer interface has a default implementation of Dispose() that calls GetScopes() and disposes each scope. Your explicit IDisposable.Dispose() implementation also disposes all scopes. While you check _disposed in GetScopes() and return empty array (line 33-36), this coordination is fragile.

Suggested fix: Consider documenting this coordination or restructuring to make it clearer.

### 3. Missing ConfigureAwait(false) on AsyncServiceScope disposal (line 59)
The cast to IAsyncDisposable should be followed by ConfigureAwait(false) for consistency with the rest of the codebase's async patterns.

## Verdict
⚠️ REQUEST CHANGES - Critical race condition in issue #1

PR #1794, opened 2026-01-02
perf(engine): Cache module attribute lookups to reduce reflection overhead (#1483)

### LINKED ISSUE #1483: Performance: Uncached GetCustomAttribute calls in ModuleScheduler.InitializeModules
## Issue Location
- **File**: `src/ModularPipelines/Engine/ModuleScheduler.cs`
- **Lines**: 97, 118, 129

## Problem
The `InitializeModules` method calls `GetCustomAttribute` for each module without caching:

```csharp
foreach (var state in _moduleStates.Values)
{
    var moduleType = state.ModuleType;
    
    var notInParallelAttr = moduleType.GetCustomAttribute<NotInParallelAttribute>();
    // ... use attribute ...
    
    var priorityAttr = moduleType.GetCustomAttribute<PriorityAttribute>();
    // ... use attribute ...
    
    var executionHintAttr = moduleType.GetCustomAttribute<ExecutionHintAttribute>();
    // ... use attribute ...
}
```

While .NET does cache reflection metadata internally, there's still overhead from:
1. The reflection lookup each time
2. Attribute construction (if not cached)
3. Type hierarchy traversal

This happens during module initialization which is a startup cost multiplied by the number of modules.

## Suggested Fix
Consider caching module metadata in a dedicated cache class similar to `ModuleAttributeEventService`:

```csharp
internal class ModuleMetadataCache
{
    private static readonly ConcurrentDictionary<Type, ModuleSchedulingMetadata> Cache = new();
    
    public static ModuleSchedulingMetadata GetMetadata(Type moduleType)
    {
        return Cache.GetOrAdd(moduleType, static t => new ModuleSchedulingMetadata
        {
            NotInParallelAttribute = t.GetCustomAttribute<NotInParallelAttribute>(),
            PriorityAttribute = t.GetCustomAttribute<PriorityAttribute>(),
            ExecutionHintAttribute = t.GetCustomAttribute<ExecutionHintAttribute>()
        });
    }
}

internal record ModuleSchedulingMetadata
{
    public NotInParallelAttribute? NotInParallelAttribute { get; init; }
    public PriorityAttribute? PriorityAttribute { get; init; }
    public ExecutionHintAttribute? ExecutionHintAttribute { get; init; }
}
```

## Related Files
Similar patterns exist in:
- `ModuleConditionHandler.cs` (lines 55, 69, 76-77)
- `ModuleRunner.cs` (line 189)

## Impact
Low to Medium - affects pipeline startup time, especially noticeable with large numbers of modules.

thomhurst:
## Summary
- Create `ModuleMetadataCache` with static `ConcurrentDictionary` for caching
- Create `ModuleSchedulingMetadata` record to hold cached attributes
- Update `ModuleScheduler` to use cached metadata instead of repeated reflection

Fixes #1483

## Test Plan
- [x] Build succeeds
- [ ] Unit tests pass

### DIFF
--- a/src/ModularPipelines/Engine/ModuleMetadataCache.cs
+++ b/src/ModularPipelines/Engine/ModuleMetadataCache.cs
@@ -1,0 +1,28 @@
+using System.Collections.Concurrent;
+using System.Reflection;
+using ModularPipelines.Attributes;
+
+namespace ModularPipelines.Engine;
+
+/// <summary>
+/// Caches module scheduling metadata to avoid repeated reflection lookups.
+/// </summary>
+internal static class ModuleMetadataCache
+{
+    private static readonly ConcurrentDictionary<Type, ModuleSchedulingMetadata> Cache = new();
+
+    /// <summary>
+    /// Gets the cached scheduling metadata for a module type.
+    /// </summary>
+    /// <param name="moduleType">The module type to get metadata for.</param>
+    /// <returns>The cached metadata containing scheduling attributes.</returns>
+    public static ModuleSchedulingMetadata GetMetadata(Type moduleType)
+    {
+        return Cache.GetOrAdd(moduleType, static t => new ModuleSchedulingMetadata
+        {
+            NotInParallelAttribute = t.GetCustomAttribute<NotInParallelAttribute>(),
+            PriorityAttribute = t.GetCustomAttribute<PriorityAttribute>(),
+            ExecutionHintAttribute = t.GetCustomAttribute<ExecutionHintAttribute>(),
+        });
+    }
+}
--- a/src/ModularPipelines/Engine/ModuleScheduler.cs
+++ b/src/ModularPipelines/Engine/ModuleScheduler.cs
@@ -97,10 +97,12 @@ internal class ModuleScheduler : IModuleScheduler
         {
             var moduleType = state.ModuleType;
 
-            var notInParallelAttr = moduleType.GetCustomAttribute<NotInParallelAttribute>();
-            if (notInParallelAttr != null)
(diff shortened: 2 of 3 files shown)


thomhurst:
## Summary
This PR introduces caching for module attribute lookups to reduce reflection overhead during module initialization.

## Critical Issues
None found ✅

## Suggestions

### 1. Consider thread-safety documentation for static cache
The static `ConcurrentDictionary` in `ModuleMetadataCache` is thread-safe for concurrent reads/writes, but it would be helpful to add a comment clarifying the lifetime of this cache (application-wide, persists across pipeline runs). This could be relevant for scenarios where module types might be unloaded or reloaded.

**src/ModularPipelines/Engine/ModuleMetadataCache.cs:11**
```csharp
// Application-wide cache - persists across pipeline runs
// Thread-safe for concurrent access via ConcurrentDictionary
private static readonly ConcurrentDictionary<Type, ModuleSchedulingMetadata> Cache = new();
```

### 2. Consider caching in AddModule method (future enhancement)
The `AddModule` method (ModuleScheduler.cs:171-229) is used for dynamically adding modules (e.g., SubModules). While it doesn't directly look up the scheduling attributes that are now cached, this method could benefit from the same caching pattern if it's used frequently in hot paths. However, this may be out of scope for this PR since the issue #1483 specifically targets `InitializeModules`.

## Verdict

The implementation correctly:
- Uses `ConcurrentDictionary.GetOrAdd` with a static lambda to avoid repeated reflection
- Caches only immutable attribute data (attributes can't change after type load)
- Maintains the same behavior as the original code
- Addresses the performance concern raised in #1483

This is a clean, focused performance improvement with minimal risk.

PR #1802, opened 2026-01-02
perf: Cache reflection in PipelineHost.DisposeAsync

### LINKED ISSUE #1473: Performance: PipelineHost.DisposeAsync uses reflection on every call
## Issue Location
- **File**: - **Lines**: 75-77

## Problem
The \ method uses reflection to access the internal \ property on every disposal:

\
Issues:
1. \ and \ are called every time \ is invoked
2. Accessing internal implementation details of the DI container is fragile
3. No caching of the PropertyInfo

## Suggested Fix
Option 1: Cache the PropertyInfo as a static field:
\
Option 2: Use a compiled delegate for property access (following the pattern in \):
\
Option 3: Consider if this reflection is even needed - perhaps the IAsyncDisposable pattern on the scope is sufficient.

## Impact
Low - disposal typically happens once, but it's a good practice to avoid reflection in hot paths and this creates cleaner code.

thomhurst:
## Summary
- Cache the `PropertyInfo` lookup for the `Disposables` property using a `ConcurrentDictionary<Type, PropertyInfo?>` keyed by service provider type
- Avoids repeated reflection calls on every `DisposeAsync` invocation
- Uses `static` lambda to prevent closure allocation

Fixes #1473

## Test plan
- [x] Build passes (`dotnet build`)
- [ ] Existing tests continue to pass
- [ ] No behavior change, just performance improvement

### DIFF
--- a/src/ModularPipelines/Host/PipelineHost.cs
+++ b/src/ModularPipelines/Host/PipelineHost.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
@@ -14,6 +15,8 @@ namespace ModularPipelines.Host;
 /// </summary>
 internal class PipelineHost : IPipelineHost
 {
+    private static readonly ConcurrentDictionary<Type, PropertyInfo?> DisposablesPropertyCache = new();
+
     private readonly IHost _hostImplementation;
     private readonly AsyncServiceScope _serviceScope;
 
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Caches reflection lookup for the service provider's internal Disposables property to improve performance on repeated pipeline executions.

## Critical Issues
None found ✅

## Suggestions

**Consider thread-safety of the cached PropertyInfo usage**: While the ConcurrentDictionary safely caches the PropertyInfo lookup, the actual reflection access (GetValue(Services)) happens outside the cache. This is fine since PropertyInfo.GetValue is thread-safe, but it's worth noting that the Services instance could theoretically change between calls if service provider types differ across invocations.

**Reflection on internal implementation details**: This code relies on accessing a private "Disposables" property of the service provider implementation. While the caching improves performance, the underlying approach is fragile - if the Microsoft.Extensions.DependencyInjection internals change, this breaks. Consider:
- Is there a public API to enumerate disposables?
- Could this be handled through a custom service scope implementation instead?
- Worth adding a comment explaining why this reflection is necessary

These are minor observations - the caching implementation itself is solid and the static lambda prevents allocations as intended.

## Verdict
✅ **APPROVE** - No critical issues, solid performance optimization

PR #1803, opened 2026-01-02
perf: Cache MakeGenericType in ModuleRunner.GetOrCreateLogger

### LINKED ISSUE #1478: Performance: ModuleRunner.GetOrCreateLogger uses MakeGenericType on every call
## Issue Location
- **File**: `src/ModularPipelines/Engine/Execution/ModuleRunner.cs`
- **Lines**: 280-285

## Problem
The `GetOrCreateLogger` method calls `MakeGenericType` on every module execution:

```csharp
private IModuleLogger GetOrCreateLogger(Type moduleType, IServiceProvider scopedServiceProvider)
{
    var loggerType = typeof(ModuleLogger<>).MakeGenericType(moduleType);
    return _loggerContainer.GetLogger(loggerType)
        ?? (IModuleLogger)scopedServiceProvider.GetRequiredService(loggerType);
}
```

While `MakeGenericType` is relatively fast, it still involves:
1. Type validation
2. Generic argument checking  
3. Internal caching lookup

This method is called for every module execution, making it a hot path.

## Suggested Fix
Cache the constructed generic type in a ConcurrentDictionary:

```csharp
private static readonly ConcurrentDictionary<Type, Type> LoggerTypeCache = new();

private IModuleLogger GetOrCreateLogger(Type moduleType, IServiceProvider scopedServiceProvider)
{
    var loggerType = LoggerTypeCache.GetOrAdd(moduleType, 
        static mt => typeof(ModuleLogger<>).MakeGenericType(mt));
    return _loggerContainer.GetLogger(loggerType)
        ?? (IModuleLogger)scopedServiceProvider.GetRequiredService(loggerType);
}
```

## Impact
Low to Medium - this is called for every module execution, so caching provides consistent performance improvement, especially for pipelines with many module executions.

thomhurst:
## Summary
- Add a static `ConcurrentDictionary<Type, Type>` cache in `IModuleLoggerContainer` to store the results of `MakeGenericType` calls
- The `GetOrCreateLogger` method now uses `GetOrAdd` to retrieve cached logger types instead of calling `MakeGenericType` on every module execution
- Uses a `static` lambda to avoid allocations in the hot path

## Why
`MakeGenericType` is relatively expensive due to runtime type construction. Since each module type maps to a single logger type, caching this relationship eliminates redundant work during pipeline execution.

Fixes #1478

## Test plan
- [x] Build passes (`dotnet build`)
- [ ] Existing tests continue to pass (CI verification)

### DIFF
--- a/src/ModularPipelines/Logging/IModuleLoggerContainer.cs
+++ b/src/ModularPipelines/Logging/IModuleLoggerContainer.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace ModularPipelines.Logging;
@@ -8,6 +9,12 @@ namespace ModularPipelines.Logging;
 internal interface IModuleLoggerContainer
 {
     /// <summary>
+    /// Cache for constructed generic logger types.
+    /// MakeGenericType is expensive, so we cache the result per module type.
+    /// </summary>
+    private static readonly ConcurrentDictionary<Type, Type> LoggerTypeCache = new();
+
+    /// <summary>
     /// Gets an existing logger or creates one from the scoped service provider.
     /// </summary>
     /// <param name="moduleType">The module type to get/create a logger for.</param>
@@ -15,7 +22,10 @@ internal interface IModuleLoggerContainer
     /// <returns>The module logger instance.</returns>
     IModuleLogger GetOrCreateLogger(Type moduleType, IServiceProvider scopedServiceProvider)
     {
-        var loggerType = typeof(ModuleLogger<>).MakeGenericType(moduleType);
+        var loggerType = LoggerTypeCache.GetOrAdd(
+            moduleType,
+            static type => typeof(ModuleLogger<>).MakeGenericType(type));
+
         return GetLogger(loggerType)
             ?? (IModuleLogger)scopedServiceProvider.GetRequiredService(loggerType);
     }


thomhurst:
## Summary
Adds a static cache to avoid repeated `MakeGenericType` calls in the module logger creation hot path.

## Critical Issues
None found ✅

## Suggestions

**Consider extracting cache to the implementation class**

While the current approach (static field in interface default method) is technically valid in C# 8.0+, it's unconventional. Consider moving the cache to `ModuleLoggerContainer` instead:

```csharp
// In ModuleLoggerContainer.cs
private static readonly ConcurrentDictionary<Type, Type> LoggerTypeCache = new();

// In IModuleLoggerContainer.cs - remove the static field and update:
IModuleLogger GetOrCreateLogger(Type moduleType, IServiceProvider scopedServiceProvider)
{
    var loggerType = ModuleLoggerContainer.GetCachedLoggerType(moduleType);
    return GetLogger(loggerType)
        ?? (IModuleLogger)scopedServiceProvider.GetRequiredService(loggerType);
}

// Add static method to ModuleLoggerContainer
internal static Type GetCachedLoggerType(Type moduleType) =>
    LoggerTypeCache.GetOrAdd(moduleType, static type => typeof(ModuleLogger<>).MakeGenericType(type));
```

This would:
- Keep implementation details in the concrete class
- Make the code more discoverable (static state is easier to spot in classes than in interface default methods)
- Follow conventional C# patterns

However, this is a minor preference - the current implementation is functionally correct and achieves the performance goal.

## Verdict

The performance optimization is sound and correctly implemented. The use of `ConcurrentDictionary.GetOrAdd` with a static lambda avoids allocations in the hot path, and caching `MakeGenericType` results is a valid optimization for this use case.

Issue #1812, opened 2026-01-02
Refactor: Logging, Progress, and Results Infrastructure Design Improvements


thomhurst:
## Summary

A thorough analysis of the logging, progress printing, and results printing infrastructure has identified several design concerns that violate SOLID, KISS, SRP, and DRY principles. While the current implementation is functional, addressing these issues will improve maintainability, testability, and reduce complexity.

## Analysis Scope

Files analyzed in `src/ModularPipelines/`:
- `Logging/` - 25 files including ModuleLogger, ModuleOutputWriter, CommandLogger, LogEventBuffer
- `Helpers/` - ProgressPrinter, ConsolePrinter, MarkupFormatter, StatusFormatter, TimeSpanFormatter
- `Engine/Executors/` - PipelineOutputCoordinator, ExecutionOrchestrator
- `Engine/BuildSystemFormatters/` - 9 CI/CD-specific formatters

---

#### Issue A: ProgressPrinter has dual responsibilities
- **File**: `src/ModularPipelines/Helpers/ProgressPrinter.cs`
- Handles BOTH real-time progress display AND final results table printing
- `PrintProgress()` vs `PrintResults()` are fundamentally different concerns
- Methods span 373 lines combining progress tracking with tabular result formatting

#### Issue B: ModuleLogger<T> implements too many interfaces
- **File**: `src/ModularPipelines/Logging/ModuleLogger.cs`
- Implements `IModuleLogger`, `ILogger<T>`, `IConsoleWriter`, `IDisposable`
- Mixes logging, console output, and lifecycle management

#### Issue C: ModuleOutputWriter handles multiple concerns
- **File**: `src/ModularPipelines/Logging/ModuleOutputWriter.cs`
- Buffering, CI-specific formatting, section headers, secret obfuscation, flushing
- Contains nested `OutputSection` class adding complexity

---

#### Issue A: Repeated timestamp/duration formatting
- `CommandLogger.GetTimestampPrefix()` at line 212
- `ProgressPrinter.GetTime()` at line 362
- `ModuleOutputContext.FormatSectionHeader()`
- `TimeSpanFormatter.ToDisplayString()` is available but not consistently used

#### Issue B: Duplicated progress task handling
- `Handle(ModuleStartedNotification)` and `Handle(SubModuleCreatedNotification)` in `ProgressPrinter.cs:75-241`
- Nearly identical progress ticking logic repeated in both handlers
- Same 10% headroom calculation, same ticking loop structure

[...]

PR #1823, opened 2026-01-02
refactor: Logging infrastructure thread-safety and modernization (#1812)

thomhurst:
## Summary

This PR addresses several items from issue #1812:

- **Add thread-safety to AfterPipelineLogger** - Multiple modules can call `LogOnPipelineEnd()` concurrently
- **Replace legacy MethodImplOptions.Synchronized** with modern lock pattern in ModuleLoggerProvider

### 1. Thread-Safety Gap - AfterPipelineLogger (Fixed)

Added thread-safety using explicit `lock` synchronization:
- `LogOnPipelineEnd()`: Protected `_values.Add()` with lock
- `GetOutput()`: Protected entire method body with lock
- `WriteLogs()`: Created a copy of `_values` under lock before iterating

### 2. Legacy Synchronization Pattern (Fixed)

Replaced `[MethodImpl(MethodImplOptions.Synchronized)]` in ModuleLoggerProvider with explicit `lock` statement:
- Removed legacy `System.Runtime.CompilerServices` usage
- Added private `_lock` object for better encapsulation
- The legacy pattern locks on `this` which can cause deadlocks if external code acquires the same lock

### 3. DRY Investigation (No changes needed)

Investigated duration formatting - the codebase already consistently uses `TimeSpanFormatter.ToDisplayString()`.

## Test plan
- [x] Build passes
- [ ] Existing tests pass
- [ ] Verify thread-safety under concurrent access

### DIFF
--- a/src/ModularPipelines/Logging/AfterPipelineLogger.cs
+++ b/src/ModularPipelines/Logging/AfterPipelineLogger.cs
@@ -13,6 +13,7 @@ internal class AfterPipelineLogger : IAfterPipelineLogger
     private readonly StringBuilder _stringBuilder = new();
     private readonly List<string> _values = [];
     private readonly object _lock = new();
+    private bool _isCacheValid;
 
     public AfterPipelineLogger(ILogger<AfterPipelineLogger> logger)
     {
@@ -27,8 +28,7 @@ internal class AfterPipelineLogger : IAfterPipelineLogger
         lock (_lock)
         {
             _values.Add(value);
-            // Clear cached output so GetOutput() rebuilds it with the new value
-            _stringBuilder.Clear();
+            _isCacheValid = false;
         }
     }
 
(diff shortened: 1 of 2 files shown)


thomhurst:
## Additional Fixes Applied (Commit 949233ff3)

Addressed the issues from the second review:

### 1. AfterPipelineLogger: Fixed caching logic
- Now uses a boolean `_isCacheValid` flag instead of relying on `StringBuilder.Length`
- `LogOnPipelineEnd()` sets `_isCacheValid = false`
- `GetOutput()` checks the flag, clears StringBuilder and rebuilds if invalid, then sets flag to true
- Prevents unnecessary rebuilds when GetOutput() is called multiple times with no new values

### 2. ModuleLoggerProvider: Fixed caching separation
- `GetLogger(Type type)` no longer uses or modifies `_moduleLogger`
- Removed lock from `GetLogger(Type type)` since it doesn't share the cache
- Only parameterless `GetLogger()` caches to `_moduleLogger` (with lock protection)
- Avoids potential deadlock - `GetLogger(Type type)` doesn't hold lock during DI resolution
- Renamed `MakeLogger` to `CreateLogger` for clarity

Issue #1813, opened 2026-01-02
Refactor: Module System Design - ISP/SRP Violations and Code Duplication


thomhurst:
## Summary

The Module System in ModularPipelines has several design issues that violate SOLID principles, particularly around interface segregation, single responsibility, and code duplication. This issue documents the findings and recommends refactoring approaches.

### 1. Interface Segregation Violation - Behavior Interface Pattern (CRITICAL)

**Location**: `src/ModularPipelines/Modules/Behaviors/*.cs`

The module system uses marker interfaces for behaviors (`ISkippable`, `IRetryable`, `ITimeoutable`, `IIgnoreFailures`, `IHookable`, `IAlwaysRun`) with runtime type checking in `ModuleExecutionPipeline.cs`.

**Problem**: Sequential type checks using `if (module is ISkippable)`, `if (module is IHookable)`, etc. repeated 7+ times throughout execution logic (lines 76-82, 89-92, 132-142, 155, 241-246, 253-256, 321-334).

```csharp
// ModuleExecutionPipeline.cs, lines 76-92
if (module is ISkippable skippable)
{
    var skipDecision = await skippable.ShouldSkip(moduleContext).ConfigureAwait(false);
    // ...
}

if (module is IHookable hookable)
{
    await hookable.OnBeforeExecute(moduleContext).ConfigureAwait(false);
}
```

**Impact**:
- No unified behavior protocol
- Repeated type checking is inefficient
- Adding new behaviors requires changes in multiple locations

**Recommendation**: Replace marker interface pattern with a behavior builder/composition pattern or use a strategy pattern with registered behaviors.

---

### 2. SRP Violation - ModuleExecutionContext Overloaded (CRITICAL)

**Location**: `src/ModularPipelines/Engine/ModuleExecutionContext.cs` (lines 19-254)

**Multiple Responsibilities**:
1. Execution State Tracking (lines 45-70): Status, StartTime, EndTime, Duration
2. Cancellation Management (lines 79-84, 163-165): ModuleCancellationTokenSource
3. Result Management (lines 21, 106-125): TaskCompletionSource
4. Timing/Metrics (lines 75, 147-151): Stopwatch, RecordEndTime()
5. Sub-module Tracking (lines 29, 89): SubModules list management
6. Type System Management (lines 40, 162): ModuleType, TypedModule handling

**Recommendation**: Split into focused classes:
- `ModuleExecutionState` - status and state tracking
[...]

PR #1827, opened 2026-01-02
refactor(engine): Add behavior metadata caching infrastructure and simplify AlwaysRun check (#1813)

thomhurst:
## Summary

- Add `ModuleBehaviorMetadata` class with type-safe interface caching infrastructure
- Expand `ModuleMetadataCache` with `GetBehaviorMetadata` for future optimization
- Rename `GetMetadata` to `GetSchedulingMetadata` for clarity
- Simplify AlwaysRun check to use `ModuleRunType` property directly

### ModuleBehaviorMetadata (Infrastructure for Future Optimization)
Added caching infrastructure for behavior interface checks. This provides:
- Type-safe `IsAssignableFrom` checks (not string-based name matching)
- Cached results via `ConcurrentDictionary`
- Ready for integration when optimizing `ModuleExecutionPipeline`

**Note:** This PR adds the infrastructure only. Integrating it into `ModuleExecutionPipeline` 
(replacing the `is` checks with cached lookups) would be a separate, larger refactoring effort.
The infrastructure is tested and ready for that future work.

### Simplified AlwaysRun Check
Removed redundant `module is IAlwaysRun` check since `ModuleRunType` property already 
performs this check internally (see `IModule.cs:21`).

## What's NOT Included (Deferred)
- `ExecutionTimingTracker` - Removed as it requires deeper integration with 
  `ModuleExecutionContext` and `SubModule` (future work)
- Replacing `is` checks in `ModuleExecutionPipeline` with cached metadata lookups 
  (requires careful testing, deferred to avoid scope creep)

## Test plan
- [x] Build succeeds with no errors
- [x] Module-related unit tests pass

Partially addresses #1813 (behavior metadata caching infrastructure)


thomhurst:
## Summary
Refactors module metadata caching by adding ExecutionTimingTracker and ModuleBehaviorMetadata classes, and simplifies AlwaysRun detection in ModuleExecutionPipeline.

## Critical Issues

### 1. New classes are not used anywhere
- ExecutionTimingTracker is defined but never instantiated or used
- ModuleBehaviorMetadata is defined but GetBehaviorMetadata() is never called
- These appear to be prepared for future use but don't serve any purpose in this PR

**Impact**: Dead code that adds complexity without benefit. The PR description mentions consolidating timing logic from ModuleExecutionContext and SubModule, but those classes are not modified to use ExecutionTimingTracker.

**Recommendation**: Either use these classes in this PR or remove them and introduce them when actually needed.

### 2. String-based interface detection is fragile
In ModuleBehaviorMetadata.FromType() lines 56-67, the code matches interfaces by simple name only (i.Name), ignoring namespace. This means:
- If someone creates MyNamespace.ISkippable, it will falsely match
- StartsWith(IRetryable) could match unintended interfaces

**Correct approach** used elsewhere in the codebase is typeof(ISkippable).IsAssignableFrom(moduleType) for proper type hierarchy checks. For generic IRetryable check i.GetGenericTypeDefinition() == typeof(IRetryable). This avoids namespace collisions.

### 3. Obsolete attribute on GetMetadata creates confusion
Line 48 marks GetMetadata as obsolete immediately after renaming it. Only one call site is updated (ModuleScheduler.cs line 116). Either fully rename all call sites in this PR or do not mark it obsolete yet.

## Suggestions

ModuleExecutionPipeline.cs line 154 comment says Check both the interface and the property but the code was changed to only check ModuleRunType property which internally checks IAlwaysRun interface per IModule.cs line 21. Update the comment.

## Verdict
REQUEST CHANGES - The PR introduces unused code and uses fragile string-based interface detection that could cause runtime bugs. The simplification to ModuleExecutionPipeline is good but the new infrastructure needs to either be used or removed.

thomhurst:
## Summary
This PR adds behavior metadata caching infrastructure and simplifies the AlwaysRun check.

## Critical Issues

### 1. PR Description Mismatch - Missing ExecutionTimingTracker
The PR description claims to add `ExecutionTimingTracker` class to consolidate timing logic (addressing issue #1813, item #3), but this class is **not present in the diff**. The PR only contains:
- `ModuleBehaviorMetadata` class
- Cache infrastructure in `ModuleMetadataCache`
- Simplified AlwaysRun check
- Renamed `GetMetadata` to `GetSchedulingMetadata`

Either the PR description is incorrect or changes are missing.

### 2. Unused Code - ModuleBehaviorMetadata Not Utilized
The PR adds `ModuleBehaviorMetadata` caching infrastructure but **never uses it**. 

Current code in `ModuleExecutionPipeline.cs` still performs runtime type checks:
- Line 76: `if (module is ISkippable skippable)`
- Line 89: `if (module is IHookable hookable)`
- Line 132: `if (module is IHookable hookable)`
- Line 241: `if (module is ITimeoutable timeoutable)`
- Line 321: `if (module is IIgnoreFailures ignoreFailures)`

The new `GetBehaviorMetadata()` method is never called. This adds ~73 lines of unused code to the codebase, which violates the DRY principle the PR claims to improve.

**Expected**: Replace the `is` checks with cached metadata lookups.

### 3. Incomplete Refactoring
If this is part 1 of a multi-part refactoring, it should be clearly stated. Adding infrastructure without usage in the same PR makes it difficult to verify correctness and creates dead code.

## Suggestions

### AlwaysRun Simplification - Verify Comment Accuracy ✓
The simplification at `ModuleExecutionPipeline.cs:154` is correct since `ModuleRunType` property already checks `this is IAlwaysRun` (see `IModule.cs:21`). However, the comment says:

```csharp
// ModuleRunType property already checks for IAlwaysRun interface (see IModule.cs:21)
```

While technically accurate, the comment could be clearer about WHY this matters - it prevents a redundant interface check.

## Verdict
⚠️ **REQUEST CHANGES** - Critical issues found

The PR contains unused code and doesn't match its description. Either:
1. Add the missing `ExecutionTimingTracker` class and update the code to use `ModuleBehaviorMetadata`, OR
2. Update the PR description to accurately reflect what's included and explain why the infrastructure is added without usage

thomhurst:
## Response to Review Comments

Updated the PR title and description to accurately reflect what's included.

### Regarding the Unused Infrastructure

You're right that `ModuleBehaviorMetadata` is not currently used. This is intentional - it's **infrastructure prep** for a future optimization.

**Why add it now?**
- The type-safe interface checking pattern is valuable and tested
- It documents the intended optimization path from issue #1813
- Integrating it into `ModuleExecutionPipeline` (replacing 6+ `is` checks) is a separate refactoring with its own risks

**What would integration look like?**
```csharp
// Instead of:
if (module is ISkippable skippable)

// Could become:
var behavior = ModuleMetadataCache.GetBehaviorMetadata(executionContext.ModuleType);
if (behavior.IsSkippable)
{
    var skippable = (ISkippable)module;
    // ...
}
```

However, this changes the execution pattern and requires careful testing to ensure no regressions. I'd prefer to do that in a dedicated PR.

### Options for Moving Forward

1. **Keep as infrastructure** - Merge as-is with updated description (current approach)
2. **Remove until needed** - Delete `ModuleBehaviorMetadata` and add it when actually integrating
3. **Full integration** - Replace the `is` checks now (larger scope change)

Happy to go with option 2 if you prefer not having unused infrastructure.

thomhurst:
Closing this PR - the behavior metadata caching is a micro-optimization that won't provide meaningful performance improvement. The `is` checks happen once per module and take nanoseconds, while modules run for seconds to minutes doing actual work.

The AlwaysRun simplification was already merged as part of the main branch updates.

PR #1886, opened 2026-01-05
feat(validation): Add early module dependency validation at host creation time

### LINKED ISSUE #1877: Self-Referencing Validation: Validate dependencies at registration time, not runtime
## Problem
Module dependency validation happens at execution time, not registration time, causing:

- **Late Error Detection**: Dependencies discovered at pipeline run time
- **Runtime Failures**: Pipeline crashes mid-execution when dependency missing
- **No Build-Time Verification**: Can't verify configuration before running
- **Poor Developer Experience**: Errors come from users, not during development
- **No Circular Dependency Detection**: Circular dependencies only caught at execution

## Current Design
```csharp
// Dependency errors not caught until pipeline execution
var builder = PipelineHostBuilder.Create()
    .AddModule<MyModuleWithMissingDependency>();

var result = await builder.BuildAndRunAsync(); 
// ERROR: Fails here at runtime if dependency not registered
```

## Proposed Solution
Validate dependencies during host building:

```csharp
public class PipelineHostBuilder
{
    public async Task<IPipelineHost> BuildAsync()
    {
        // Validate all registered modules
        var validator = new ModuleDependencyValidator(_registeredModules);
        var issues = validator.ValidateDependencies();
        
        if (issues.HasErrors)
        {
            throw new InvalidModuleConfigurationException(
                $"Module configuration errors:\n{string.Join("\n", issues)}");
        }
        
        return new PipelineHost(...);
    }
}

// Usage catches errors early
var host = await builder.BuildAsync(); // ERROR here during build, not execution
await host.RunAsync();
```

## Validation Checks
- Detect missing dependencies
- Detect circular dependencies  
- Detect invalid dependency configurations
- Validate module type compatibility
- Check for ambiguous registrations

## Benefits
- Errors caught early during development
- Clear configuration validation
- No runtime surprises
- Circular dependencies prevented
- Better error messages with context
- Enables better tooling/analysis

## Priority: MEDIUM - Improves reliability and developer experience

thomhurst:
## Summary

- Added early validation of module dependencies during pipeline host building
- Catches configuration errors before pipeline execution starts
- Detects self-references, missing dependencies, and circular dependency chains

## Changes

- Added `InvalidModuleConfigurationException` for validation errors
- Added `ModuleDependencyValidator` with comprehensive validation checks:
  - Self-referencing modules detection
  - Missing (unregistered) dependency detection  
  - Circular dependency chain detection with full cycle reporting
- Integrated validation into `PipelineHost.Create()` before `InitializeAsync()`

## Test plan

- [ ] Verify existing tests pass
- [ ] Test with intentionally misconfigured modules
- [ ] Verify clear error messages for each validation type

Closes #1877


thomhurst:
## Summary
Adds early validation of module dependencies at host creation time to catch configuration errors before pipeline execution.

## Critical Issues

### 1. Missing Test Coverage
The PR adds a new `ModuleDependencyValidator` class with 198 lines of validation logic, but there are no tests included. Given that this is a critical path that throws exceptions, comprehensive tests are essential:
- Tests for self-reference detection
- Tests for missing dependency detection  
- Tests for circular dependency detection (simple and nested)
- Tests for `IgnoreIfNotRegistered` flag behavior
- Tests with `DependsOnAllModulesInheritingFromAttribute`

The existing tests in `DirectCollisionTests.cs` and `NestedCollisionTests.cs` validate the old `DependencyCollisionDetector` path. New tests should verify that the early validation in `PipelineHost.Create()` catches these same issues.

### 2. Potential Duplicate Validation
The codebase already has `DependencyCollisionDetector` (src/ModularPipelines/Helpers/DependencyCollisionDetector.cs:21-60) that performs circular dependency detection. This is called during pipeline initialization via `DependencyDetector.Check()` (src/ModularPipelines/Engine/Executors/PipelineInitializer.cs:57).

The new `ModuleDependencyValidator` appears to duplicate this validation at an earlier stage. While fail-fast is good, this creates maintenance burden:
- Two separate implementations of cycle detection algorithms
- Two places to maintain the same logic
- Different error message formats (though both use the same exception types)

**Suggestion**: Consider whether the old `DependencyCollisionDetector` should be removed or if the new validator should reuse existing logic. If both are intentionally kept, add a comment explaining why duplicate validation is necessary.

### 3. Algorithm Difference in Self-Reference Check  
In `ModuleDependencyValidator.cs` (new code, line 44-58):
```csharp
var dependencies = ModuleDependencyResolver.GetDependencies(moduleType, moduleTypes);
foreach (var (dependencyType, _) in dependencies)
{
    if (dependencyType == moduleType)
    {
        throw new ModuleReferencingSelfException(...);
    }
}
```

In `DependencyCollisionDetector.cs` (existing code, line 32-38):
```csharp
if (moduleDependencyModel.IsDependentOn.Any(d => d.Module.GetType() == moduleType))
{
    throw new ModuleReferencingSelfException(...);
}
```

Both check for self-reference, but in different ways. This inconsistency could lead to different behavior if `ModuleDependencyResolver.GetDependencies` filters or transforms dependencies differently than `moduleDependencyModel.IsDependentOn`.

## Suggestions

### 1. Consider Timing of Validation
The new validation runs in `PipelineHost.Create()` after `hostBuilder.Build()` but **before** `InitializeAsync()`. This is earlier than the old validation which runs during `PipelineInitializer` after services are fully initialized.

**Question**: Does this early timing handle all edge cases? Specifically:
- Dynamic dependencies added via `IModuleDependencyRegistry` 
- SubModules discovered during execution
- Modules added via lifecycle events

The validator only sees modules registered at host creation time. If your intent is to catch obvious configuration errors early, this is fine—but document this limitation.

### 2. Performance Consideration
The new validator runs a DFS-based cycle detection on every host creation. For pipelines with hundreds of modules, this could add noticeable startup time. Consider:
- Adding performance benchmarks
- Providing an option to disable validation in production (if startup time is critical)
- Caching validation results if the same module types are frequently reused

### 3. Code Duplication in PipelineHost.cs
The PR moves the `Services` and `RootServices` properties around unnecessarily. The only functional change is adding the `ValidateModuleDependencies` call. Consider minimizing diff noise by keeping property positions unchanged.

## Verdict
⚠️ **REQUEST CHANGES** - Missing test coverage is blocking. The duplicate validation concern should also be addressed or explained.

PR #1889, opened 2026-01-05
feat(core): Add programmatic dependency declaration support

### LINKED ISSUE #1870: Static-Only Dependency Declarations: Support dynamic and conditional dependencies
## Problem
Dependencies are declared statically via [DependsOn<T>] attributes at compile-time only:

- **No Dynamic Dependencies**: Cannot declare dependencies based on runtime conditions
- **No Optional Dependencies**: Cannot express 'depends on this if available'
- **No Conditional Chains**: Cannot chain conditional logic in dependency resolution
- **Inflexible**: Cannot refactor based on configuration without creating new modules

## Current Design
```csharp
// Dependencies fixed at compile time
[DependsOn<DatabaseModule>]
[DependsOn<LoggingModule>]
public class ApiModule : Module<void> { }

// No way to express: 'depends on DatabaseModule only in production'
// No way to express: 'depends on CachingModule if available'
```

## Proposed Solution
Support both static and dynamic dependency declaration:

```csharp
public class ApiModule : Module<void>
{
    protected override void DeclareDependencies(IDependencyDeclaration deps)
    {
        // Static dependency
        deps.DependsOn<DatabaseModule>();
        
        // Dynamic/conditional dependency
        if (Environment.IsProduction)
        {
            deps.DependsOn<PerformanceMonitoringModule>();
        }
        
        // Optional dependency (doesn't fail if not registered)
        deps.DependsOnOptional<CachingModule>();
        
        // Lazy dependency (executed only if needed)
        deps.DependsOnLazy<HeavyProcessingModule>();
    }
}
```

## Benefits
- More flexible dependency expression
- Configuration-driven dependencies
- Optional/conditional dependencies support
- Runtime dependency decisions
- Better separation of concerns
- Cleaner module design for complex scenarios

## Priority: MEDIUM - Breaking change, enables more flexible designs

thomhurst:
## Summary
- Add support for declaring module dependencies programmatically at runtime
- Enables dynamic and conditional dependency relationships beyond static attributes
- Maintains full backward compatibility with existing `[DependsOn<T>]` attributes

## Changes
- Add `DeclareDependencies(IDependencyDeclaration deps)` virtual method on `Module<T>`
- Add `IDependencyDeclaration` interface with fluent methods:
  - `DependsOn<T>()` - required dependency
  - `DependsOnOptional<T>()` - optional dependency (ignored if not registered)
  - `DependsOnLazy<T>()` - lazy dependency (evaluated but not required)
  - `DependsOnIf<T>(bool/Func<bool>)` - conditional dependency
- Add `DependencyType` enum (Required, Optional, Lazy, Conditional)
- Update `ModuleDependencyResolver` to combine attribute and programmatic dependencies
- Update `DependencyWaiter` and `ModuleScheduler` for new dependency types
- Comprehensive unit tests (541 lines)

## Usage
```csharp
public class ApiModule : Module<string>
{
    protected override void DeclareDependencies(IDependencyDeclaration deps)
    {
        // Static dependency
        deps.DependsOn<DatabaseModule>();
        
        // Conditional dependency
        if (Environment.IsProduction)
            deps.DependsOn<PerformanceMonitoringModule>();
        
        // Optional dependency (doesn't fail if not registered)
        deps.DependsOnOptional<CachingModule>();
    }
}
```

## Test plan
- [x] Unit tests for required dependencies
- [x] Unit tests for optional dependencies
- [x] Unit tests for conditional dependencies
- [x] Unit tests for lazy dependencies
- [x] Tests for combined attribute + programmatic dependencies
- [x] Tests for dependency resolution order

Closes #1870

### DIFF
--- a/src/ModularPipelines/Modules/IDependencyDeclaration.cs
+++ b/src/ModularPipelines/Modules/IDependencyDeclaration.cs
@@ -126,27 +126,40 @@ public interface IDependencyDeclaration
     IDependencyDeclaration DependsOnIf(Type moduleType, Func<bool> predicate);
 
     /// <summary>
-    /// Declares a lazy dependency that is only executed when explicitly awaited.
+    /// Declares a lazy dependency - an optional dependency intended to be awaited on-demand.
     /// </summary>
     /// <typeparam name="TModule">The type of module to depend on.</typeparam>
     /// <returns>This instance for method chaining.</returns>
     /// <remarks>
     /// <para>
-    /// Lazy dependencies are registered as optional dependencies that do not block
-    /// the current module from starting. The lazy module will only execute if
-    /// the current module explicitly awaits it during execution.
+    /// Lazy dependencies behave the same as <see cref="DependsOnOptional{TModule}"/> for
+    /// dependency resolution purposes - the dependency is optional and will not fail if not
+    /// registered. This is a semantic marker to indicate intent that the dependency may be
+    /// awaited on-demand rather than required upfront.
     /// </para>
     /// <para>
-    /// Use this for heavy processing that may not always be needed.
+    /// <strong>Important:</strong> The lazy module will still execute during normal pipeline
+    /// scheduling if it is registered. It does NOT defer execution until explicitly awaited.
+    /// The "lazy" designation indicates the dependency relationship is optional and the result
+    /// may be consumed on-demand, but does not affect when the module runs.
+    /// </para>
+    /// <para>
+    /// Use this to express intent for optional processing that may or may not be awaited.
     /// </para>
     /// </remarks>
     IDependencyDeclaration DependsOnLazy<TModule>() where TModule : IModule;
 
     /// <summary>
-    /// Declares a lazy dependency that is only executed when explicitly awaited.
+    /// Declares a lazy dependency - an optional dependency intended to be awaited on-demand.
     /// </summary>
     /// <param name="moduleType">The type of module to depend on.</param>
     /// <returns>This instance for method chaining.</returns>
+    /// <remarks>
+    /// <para>
+    /// Lazy dependencies behave the same as <see cref="DependsOnOptional(Type)"/> for
+    /// dependency resolution purposes. See <see cref="DependsOnLazy{TModule}"/> for full details.
+    /// </para>
+    /// </remarks>
     /// <exception cref="ArgumentException">
     /// Thrown if the type does not implement <see cref="IModule"/>.
     /// </exception>


thomhurst:
## Summary
Adds programmatic dependency declaration support via `DeclareDependencies` method, enabling dynamic and conditional dependency relationships beyond static attributes.

## Critical Issues

### 1. **Misleading Documentation for Lazy Dependencies** (src/ModularPipelines/Modules/IDependencyDeclaration.cs:133-137)

The documentation for `DependsOnLazy` is misleading and contradicts the actual implementation:

**Interface documentation says:**
> "The lazy module will only execute if the current module explicitly awaits it during execution."

**But the enum documentation (DependencyType.cs:23-26) correctly states:**
> "Behaves the same as Optional for dependency resolution purposes - the dependency is optional and will not fail if not registered."

**Reality:** Looking at the implementation:
- `DeclaredDependency.Lazy()` sets `IgnoreIfNotRegistered = true` (src/ModularPipelines/Models/DeclaredDependency.cs:31)
- `GetProgrammaticDependencies` only extracts `(dep.ModuleType, dep.IgnoreIfNotRegistered)` and discards the `Kind` enum
- No mechanism exists to defer execution until the module is awaited

The lazy module will execute immediately during the normal dependency resolution, just like an optional dependency. It will NOT wait to be explicitly awaited.

**Fix Required:**
Either implement true lazy execution (defer execution until explicitly awaited) OR update the interface documentation to match the enum documentation and clarify that "lazy" is just a semantic marker indicating the dependency is optional and may be awaited if needed, but will execute immediately if registered.

## Suggestions

None - the other issues from the previous review were addressed:
- ✅ Property renamed from `DependencyType` to `ModuleType`
- ✅ Enum field renamed to `Kind`  
- ✅ Reflection caching added via `GetDeclaredDependenciesMethodCache`
- ✅ Error message syntax is correct (`IgnoreIfNotRegistered` is a property on the attribute)

## Previous Review Status

From thomhurst's review on 2026-01-05:
- **Issue #1 (Property naming):** ✅ **FIXED** - Renamed to `ModuleType` and `Kind`
- **Issue #2 (Misleading API):** ❌ **STILL OPEN** - `DependsOnLazy` documentation is still misleading (see Critical Issue #1 above)
- **Issue #3 (Reflection caching):** ✅ **FIXED** - `ConcurrentDictionary` cache added
- **Issue #4 (Thread safety docs):** Not addressed but acceptable as-is
- **Issue #5 (Error message):** Already correct in the PR

## Verdict
⚠️ **REQUEST CHANGES** - The lazy dependency documentation is misleading and contradicts the implementation. Users will expect lazy execution but will get immediate execution of optional dependencies.

thomhurst:
## Summary
This PR adds programmatic dependency declaration to ModularPipelines, allowing modules to declare dependencies dynamically at runtime via a new `DeclareDependencies` method in addition to the existing attribute-based approach.

## Critical Issues

**1. Reflection Performance Issue - Method Cache Not Checked for Null** (src/ModularPipelines/Engine/ModuleDependencyResolver.cs:87)

In `GetProgrammaticDependencies`, the cached method lookup uses `GetOrAdd` but does not properly cache null results. For modules without `GetDeclaredDependencies`, the reflection lookup happens on **every call** instead of being cached once:

```csharp
var method = GetDeclaredDependenciesMethodCache.GetOrAdd(moduleType, type =>
    type.GetMethod("GetDeclaredDependencies",
        BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public));

if (method == null)
{
    yield break;  // Early exit - null was added to cache but we exit immediately
}
```

The code does cache the null correctly with `GetOrAdd`, but since most modules won't override `DeclareDependencies`, this will store null in the cache. This is actually fine - the cache is working correctly. **This is NOT an issue - my initial read was wrong.**

**2. Potential Initialization Ordering Issue** (src/ModularPipelines/Engine/ModuleScheduler.cs:158)

In `ModuleScheduler.InitializeModules`, the code calls:

```csharp
var dependencies = ModuleDependencyResolver.GetAllDependencies(state.Module, availableModuleTypes, _dependencyRegistry);
```

This invokes `module.GetDeclaredDependencies()` which calls the virtual `DeclareDependencies(deps)` method. If a module's `DeclareDependencies` implementation tries to access module state that hasn't been initialized yet, or tries to access the scheduler/context, it could fail. 

The existing code in Module.cs shows this is called via `GetDeclaredDependencies()` which is marked `internal`, and the docs state "This method is called once during module initialization, before execution begins." However, there's no guarantee that the module's constructor has fully completed or that dependencies are available.

**Recommendation:** Add a note in the XML docs for `DeclareDependencies` warning that it's called during initialization and should not access pipeline context, other modules, or uninitialized state.

## Suggestions

**1. Caching Opportunity for Declared Dependencies** (src/ModularPipelines/Engine/ModuleDependencyResolver.cs:89)

Currently, `GetProgrammaticDependencies` calls `method.Invoke(module, null)` via reflection every time it's called. Since `DeclareDependencies` should be deterministic and called during initialization, consider caching the actual dependency list per module type (not instance), similar to how attribute-based dependencies are cached.

However, since the method is virtual and could theoretically have instance-specific behavior, caching at the type level might not be safe. The current implementation is correct but may invoke reflection multiple times per module.

**2. XML Documentation Enhancement** (src/ModularPipelines/Modules/IDependencyDeclaration.cs:145)

The documentation for `DependsOnLazy<TModule>` is thorough but could be clearer about what "lazy" means:

> Lazy dependencies behave the same as DependsOnOptional for dependency resolution purposes - the dependency is optional and will not fail if not registered. This is a semantic marker...

Consider adding a concrete example of when to use Lazy vs Optional to help users understand the distinction.

**3. Consider Adding Duplicate Detection**

The current implementation allows declaring the same dependency multiple times (e.g., once via attribute, once via `DeclareDependencies`). While this shouldn't break anything (the dependency will just be waited on once), it could indicate a configuration mistake. Consider logging a warning or deduplicating dependencies.

## Previous Review Status
Unable to access previous comments due to GitHub API token scope limitations.

## Verdict
✅ **APPROVE** - No critical issues found. The implementation follows ModularPipelines patterns correctly, includes comprehensive tests, and the suggestions are minor improvements rather than blocking issues.

The code correctly:
- Caches MethodInfo lookups to avoid repeated reflection on method resolution
- Combines attribute-based and programmatic dependencies appropriately  
- Validates module types consistently with existing patterns
- Provides clear, well-documented API with good examples
- Includes extensive unit tests covering various scenarios

PR #1895, opened 2026-01-07
feat: Implement discriminated union pattern for ModuleResult<T>

### LINKED ISSUE #1869: Exception-Based Results: Implement discriminated union Result<T, E> pattern
## Problem
ModuleResult<T> uses exceptions-as-control-flow pattern where Value property throws instead of using proper error handling:

```csharp
public sealed class ModuleResult<T?>
{
    public T? Value // Throws if not Success!
    {
        get => _result.Match(
            success => success.Value,
            failure => throw failure.Exception, // Exception as control flow
            skipped => throw new InvalidOperationException());
    }
}
```

This causes:

- **Implicit Exceptions**: Accessing Value can throw unpredictably
- **Poor Async/Await**: Can't use try-catch around await; exception comes from Value access
- **Type Unsafety**: Must know module succeeded before accessing Value
- **Poor Pattern Matching**: No discriminated union support
- **API Confusion**: Unclear when exceptions occur

## Current Design
```csharp
var result = await myModule;
var value = result.Value; // May throw!

try
{
    var value = result.Value; // Can throw anytime Value is accessed
}
catch (Exception ex) { }
```

## Proposed Solution
Implement proper Result<T, E> discriminated union:

```csharp
public abstract record ModuleResult<T>
{
    public sealed record Success(T Value) : ModuleResult<T>;
    public sealed record Failure(Exception Error) : ModuleResult<T>;
    public sealed record Skipped : ModuleResult<T>;
}

// Usage with pattern matching
var result = await myModule;
if (result is ModuleResult<string>.Success { Value: var value })
{
    // Use value safely
}
else if (result is ModuleResult<string>.Failure { Error: var error })
{
    // Handle error
}
```

## Benefits
- Type-safe result handling
- Explicit control flow
- Works with modern C# pattern matching
- Clear semantics (success, failure, skipped)
- No hidden exceptions
- Better async/await composition
- Composable with other Result types

## Priority: HIGH - Breaking change, improves type safety and error handling

thomhurst:
## Summary

Implements issue #1869 - Discriminated Union Result Pattern for `ModuleResult<T>`.

**BREAKING CHANGE**: `ModuleResult<T>.Value` no longer throws exceptions. Use pattern matching or `IsSuccess`/`ValueOrDefault` instead.

### Changes

- **Core Discriminated Union**: Rewrote `ModuleResult<T>` as an abstract record with three sealed variants:
  - `Success(T Value)` - successful execution with a value
  - `Failure(Exception Exception)` - failed execution with an exception  
  - `Skipped(SkipDecision Decision)` - skipped execution with reason

- **Safe Accessors** (no exceptions):
  - `ValueOrDefault` - returns value or default(T)
  - `ExceptionOrDefault` - returns exception or null
  - `SkipDecisionOrDefault` - returns skip decision or null

- **Quick Checks**:
  - `IsSuccess`, `IsFailure`, `IsSkipped` boolean properties

- **Pattern Matching Helpers**:
  - `Match<TResult>(onSuccess, onFailure, onSkipped)` - functional matching
  - `Switch(onSuccess, onFailure, onSkipped)` - imperative matching

- **JSON Serialization**: Full support via custom `ModuleResultJsonConverterFactory`

- **Deprecated**: `ModuleFailedException` and `ModuleSkippedException` (use pattern matching instead)

- **Deleted**: `SkippedModuleResult.cs` and `TimedOutModuleResult.cs` (replaced by variants)

### Migration Example

```csharp
// Before (throws on failure/skip):
var value = result.Value;

// After (pattern matching - recommended):
switch (result)
{
    case ModuleResult<string>.Success { Value: var value }:
        Console.WriteLine($"Got: {value}");
        break;
    case ModuleResult<string>.Failure { Exception: var ex }:
        Console.WriteLine($"Failed: {ex.Message}");
        break;
    case ModuleResult<string>.Skipped { Decision: var skip }:
        Console.WriteLine($"Skipped: {skip.Reason}");
        break;
}

// Or use safe accessors:
if (result.IsSuccess)
{
    var value = result.ValueOrDefault;
}
```

## Test plan

- [x] All unit tests pass (486 passing, matching baseline)
- [x] Full solution builds with 0 errors
- [x] JSON serialization/deserialization works correctly
- [x] History repository integration works with UsedHistory status
- [x] Pattern matching works for all three variants

Closes #1869

### DIFF
--- a/src/ModularPipelines.Examples/Modules/Azure/AssignAccessToBlobStorageModule.cs
+++ b/src/ModularPipelines.Examples/Modules/Azure/AssignAccessToBlobStorageModule.cs
@@ -21,8 +21,8 @@ public class AssignAccessToBlobStorageModule : Module<RoleAssignmentResource>
         var storageAccount = context.GetModule<ProvisionBlobStorageAccountModule, StorageAccountResource>();
 
         var roleAssignmentResource = await context.Azure().Provisioner.Security.RoleAssignment(
-            storageAccount.Value!.Id,
-            new RoleAssignmentCreateOrUpdateContent(WellKnownRoleDefinitions.BlobStorageOwnerDefinitionId, userAssignedIdentity.Value!.Data.PrincipalId!.Value)
+            storageAccount.ValueOrDefault!.Id,
+            new RoleAssignmentCreateOrUpdateContent(WellKnownRoleDefinitions.BlobStorageOwnerDefinitionId, userAssignedIdentity.ValueOrDefault!.Data.PrincipalId!.Value)
         );
 
         return roleAssignmentResource.Value;
--- a/src/ModularPipelines.Examples/Modules/Azure/ProvisionAzureFunction.cs
+++ b/src/ModularPipelines.Examples/Modules/Azure/ProvisionAzureFunction.cs
@@ -31,7 +31,7 @@ public class ProvisionAzureFunction : Module<WebSiteResource>
             {
                 Identity = new ManagedServiceIdentity(ManagedServiceIdentityType.UserAssigned)
                 {
-                    UserAssignedIdentities = { { userAssignedIdentity.Value!.Id, new UserAssignedIdentity() } },
+                    UserAssignedIdentities = { { userAssignedIdentity.ValueOrDefault!.Id, new UserAssignedIdentity() } },
                 },
                 SiteConfig = new SiteConfigProperties
                 {
@@ -40,12 +40,12 @@ public class ProvisionAzureFunction : Module<WebSiteResource>
                         new()
                         {
                             Name = "BlobStorageConnectionString",
-                            Value = storageAccount.Value!.Data.PrimaryEndpoints.BlobUri.AbsoluteUri,
+                            Value = storageAccount.ValueOrDefault!.Data.PrimaryEndpoints.BlobUri.AbsoluteUri,
                         },
                         new()
                         {
                             Name = "BlobContainerName",
-                            Value = blobContainer.Value!.Data.Name,
+                            Value = blobContainer.ValueOrDefault!.Data.Name,
                         },
                     },
                 },
--- a/src/ModularPipelines.Examples/Modules/Azure/ProvisionBlobStorageContainerModule.cs
+++ b/src/ModularPipelines.Examples/Modules/Azure/ProvisionBlobStorageContainerModule.cs
@@ -15,7 +15,7 @@ public class ProvisionBlobStorageContainerModule : Module<BlobContainerResource>
         var blobStorageAccount = context.GetModule<ProvisionBlobStorageAccountModule, StorageAccountResource>();
 
         var blobContainerProvisionResponse = await context.Azure().Provisioner.Storage.BlobContainer(
-            blobStorageAccount.Value!.Id,
+            blobStorageAccount.ValueOrDefault!.Id,
             "MyContainer",
             new BlobContainerData()
         );


thomhurst:
## Summary
This PR implements a discriminated union pattern for ModuleResult<T>, replacing exception-throwing .Value access with type-safe pattern matching using Success/Failure/Skipped variants.

## Critical Issues
None found ✅

## Suggestions

### 1. Exception Type Security Restriction May Be Too Strict
**Location:** src/ModularPipelines/Models/ModuleResult.cs:319

The ExceptionJsonConverter only allows deserializing exceptions from the `System` namespace:

```csharp
if (exceptionType \!= null &&
    typeof(Exception).IsAssignableFrom(exceptionType) &&
    (exceptionType.Namespace?.StartsWith("System", StringComparison.Ordinal) == true))
```

**Issue:** This prevents deserializing common exceptions from other namespaces like:
- `Microsoft.Data.SqlClient.SqlException`
- `Npgsql.NpgsqlException`
- Custom application exceptions

**Suggestion:** Consider using a more permissive approach:
- Allow exceptions from well-known safe assemblies (System.*, Microsoft.*)
- OR maintain a configurable allowlist of exception types
- OR document this limitation clearly for users

If the strict restriction is intentional for security, add a comment explaining the threat model.

### 2. Timing Information Loss Not Well Documented
**Location:** src/ModularPipelines/Models/ModuleResult.cs:198-206

The `GetTimingInfo` method silently returns `TimeSpan.Zero` when timing info is invalid:

```csharp
var duration = (ctx.StartTime == DateTimeOffset.MinValue || ctx.EndTime == DateTimeOffset.MinValue)
    ? TimeSpan.Zero
    : end - start;
```

**Issue:** This makes it impossible to distinguish between:
- A module that ran instantly (0ms)
- A module with missing timing information

**Suggestion:** Consider:
- Using `TimeSpan.MinValue` or nullable `TimeSpan?` to indicate invalid timing
- OR documenting this behavior prominently in the property documentation

### 3. Pattern Matching Example Could Show More Realistic Usage
**Location:** Documentation and code examples throughout

The examples show basic pattern matching, but don't demonstrate common real-world scenarios like:
- Chaining multiple module results together
- Using the `Match` helper for functional composition
- Handling partial failures in parallel module execution

**Suggestion:** Add more comprehensive examples in the documentation showing these patterns, especially since this is a breaking change and users need clear migration guidance.

## Observations (Non-Blocking)

### ✅ Good: Comprehensive Security Documentation
The ExceptionJsonConverter has excellent security documentation explaining the rationale for limiting serialized exception information. This is a model for other converters.

### ✅ Good: Consistent Refactoring
All 60+ usages of `.Value` were updated to `.ValueOrDefault`, showing thorough attention to detail across the entire codebase.

### ✅ Good: Backward Compatibility Approach
Marking `ModuleFailedException` and `ModuleSkippedException` as `[Obsolete]` instead of deleting them gives consumers a clear upgrade path.

### ✅ Good: Null Safety
The use of null-conditional operators (e.g., `changedFiles.SkipDecisionOrDefault?.ShouldSkip == true`) throughout shows careful attention to null safety.

## Verdict

This is a well-executed refactoring that meaningfully improves type safety. The suggestions above are minor enhancements that could be addressed in follow-up PRs if desired.

thomhurst:
## Response to JSON Converter Comment

The `ModuleResultJsonConverterFactory` implementation **is present** in the PR. It's defined as nested classes within `src/ModularPipelines/Models/ModuleResult.cs`:

- **`ModuleResultJsonConverterFactory`** (lines 360-374) - The factory that creates converters for `ModuleResult<T>`
- **`ModuleResultJsonConverter<T>`** (lines 379-477) - The actual converter implementation
- **`ExceptionJsonConverter`** (lines 284-357) - Handles exception serialization with security restrictions

The `[JsonConverter(typeof(ModuleResultJsonConverterFactory))]` attribute is applied to the `ModuleResult<T>` class at line 30.

The converter includes:
- Polymorphic serialization via `$type` discriminator ("Success", "Failure", "Skipped")
- Secure exception deserialization (only allows `System.*` namespace types)
- Full round-trip support for all three variants

This was implemented as nested classes rather than separate files to keep the serialization logic co-located with the model definition.

PR #2014, opened 2026-01-13
perf: Lazy-format exception messages

### LINKED ISSUE #1999: Performance: Lazy-format exception messages
## Summary

Exception messages are formatted eagerly when exceptions are created, even if the exception is never logged or its message never accessed. For exceptions with large command outputs, this creates unnecessary CPU overhead in failure paths.

## Current Behavior

**File:** `src/ModularPipelines/Exceptions/CommandException.cs` (line 49)

```csharp
public override string Message => $"""
    Input: {input}
    Error: {GetOutput(standardOutput, standardError)}
    Exit Code: {exitCode}
    """;
```

While this is a property (not constructor formatting), the string interpolation happens on every access.

## Problem

For commands with large outputs:
1. String formatting happens even if message is never logged
2. Multiple accesses format multiple times
3. Failure paths (already slow) get slower

### Option A: Lazy<string> Caching

```csharp
public class CommandException : PipelineException
{
    private readonly Lazy<string> _formattedMessage;
    
    public CommandException(string input, string standardOutput, string standardError, int exitCode)
    {
        _formattedMessage = new Lazy<string>(() => FormatMessage(input, standardOutput, standardError, exitCode));
    }
    
    public override string Message => _formattedMessage.Value;
}
```

### Option B: Override ToString() Only

Keep `Message` minimal, format details in `ToString()`:

```csharp
public override string Message => $"Command failed with exit code {exitCode}";

public override string ToString()
{
    return $"""
        {Message}
        Input: {input}
        Error: {GetOutput(standardOutput, standardError)}
        """;
}
```

### Option C: Truncated Message with Full Details Property

```csharp
public override string Message => $"Command failed: {Truncate(standardError, 200)}";

public string FullDetails => FormatFullMessage();
```

## Recommended Approach

Option A provides the best balance - message is only formatted once, on first access.

## Benchmarking

Profile exception creation with varying output sizes:
```csharp
[Params(100, 1000, 10000, 100000)]
public int OutputSize { get; set; }

[Benchmark]
public CommandException CreateException() 
    => new CommandException("cmd", new string('x', OutputSize), "", 1);
```

## Impact

- **Non-breaking**: Internal optimization
- **Improves**: Performance in failure scenarios

## Labels
- performance
- enhancement

thomhurst:
## Summary
- Modified `CommandException` to use `Lazy<string>` for deferred message formatting
- Modified `HttpResponseException` to use `Lazy<string>` for deferred message formatting
- Expensive string formatting only occurs when the exception message is actually accessed
- Reduces overhead when exceptions are caught and handled without accessing the message

Closes #1999

## Test plan
- [x] Build passes
- [x] Existing tests pass
- [ ] Exception message still formats correctly when accessed

### DIFF
--- a/src/ModularPipelines/Exceptions/CommandException.cs
+++ b/src/ModularPipelines/Exceptions/CommandException.cs
@@ -5,6 +5,8 @@ namespace ModularPipelines.Exceptions;
 /// </summary>
 public class CommandException : PipelineException
 {
+    private readonly Lazy<string> _formattedMessage;
+
     /// <summary>
     /// Initialises a new instance of the <see cref="CommandException"/> class.
     /// Initializes a new instance of the <see cref="CommandException"/> class.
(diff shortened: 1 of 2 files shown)


thomhurst:
Closing based on code review feedback.

The reviewer correctly identified a semantic issue: the base constructor receives a short message (`"Command failed with exit code {exitCode}"`) while the overridden Message property returns the full formatted message. This creates a mismatch that could confuse logging frameworks and violates the principle that the message passed to the base constructor should match what .Message returns.

Given that:
1. Exceptions are inherently rare (by design)
2. The performance benefit of lazy string formatting is minimal
3. The semantic complexity introduces potential for confusion

The cleaner solution is to keep the original immediate formatting approach.

Thank you for the thorough code review!

PR #2027, opened 2026-01-13
feat: Add type-safe GetModule extension generator (Issue #2006)

### LINKED ISSUE #2006: [v2.0] Add source generator for type-safe per-module GetModule extensions
## Summary

Extend the existing source generator to create type-safe extension methods for accessing each module's result, eliminating the need for developers to specify both type parameters.

## Current API

```csharp
// Requires knowing and specifying both types
var result = await context.GetModule<BuildModule, BuildOutput>();
```

## Proposed Generated API

For each module in the assembly, generate:

```csharp
// Auto-generated in: BuildModule.Generated.cs
public static class BuildModuleExtensions
{
    /// <summary>
    /// Gets the result of <see cref="BuildModule"/>.
    /// </summary>
    public static ModuleResult<BuildOutput> GetBuildModule(this IModuleContext context)
        => context.GetModule<BuildModule, BuildOutput>();
}
```

Usage becomes:
```csharp
var result = await context.GetBuildModule();
```

## Benefits

1. **Type inference**: No need to specify result type
2. **Discoverability**: IntelliSense shows all available modules
3. **Documentation**: Generated XML docs link to module type
4. **Compile-time safety**: Can't request wrong result type

### Source Generator Addition

Extend `ModularPipelines.SourceGenerator`:

```csharp
[Generator]
public class ModuleAccessorGenerator : ISourceGenerator
{
    public void Execute(GeneratorExecutionContext context)
    {
        // Find all types inheriting from Module<T>
        var modules = FindModuleTypes(context.Compilation);
        
        foreach (var module in modules)
        {
            var resultType = GetResultType(module);
            var source = GenerateAccessorExtension(module, resultType);
            context.AddSource($"{module.Name}Extensions.g.cs", source);
        }
    }
}
```

### Generated Code Template

```csharp
// <auto-generated />
using System.CodeDom.Compiler;
using ModularPipelines.Context;
using ModularPipelines.Models;

namespace {ModuleNamespace}
{
    [GeneratedCode("ModularPipelines.SourceGenerator", "1.0.0")]
    public static class {ModuleName}Extensions
    {
        /// <summary>
        /// Gets the result of <see cref="{ModuleName}"/>.
        /// </summary>
        /// <param name="context">The module context.</param>
        /// <returns>The module result containing <see cref="{ResultType}"/>.</returns>
        public static ModuleResult<{ResultType}> Get{ModuleName}(this IModuleContext context)
            => context.GetModule<{ModuleName}, {ResultType}>();
    }
}
```

### Naming Conventions

- `BuildModule` → `GetBuildModule()`
- `DeployToProductionModule` → `GetDeployToProductionModule()`
- Strip "Module" suffix if present: `Build` class → `GetBuild()`

## Considerations

1. **Name collisions**: Multiple modules with same name in different namespaces
2. **Visibility**: Only generate for public modules
3. **Partial classes**: Consider generating as partial for extensibility
4. **Opt-out**: Attribute to disable generation for specific modules

## Related Issues

- #1988 - Simplify Module result access API (this is the implementation approach)

## Impact

- **Non-breaking**: Additive feature, old API still works
- **Improves**: Developer experience, type safety, discoverability

## Labels
- enhancement
- source-generator
- v2.0
- developer-experience

thomhurst:
## Summary
- Add `ModuleExtensionsGenerator` source generator that creates type-safe `GetModule<T>` and `GetModuleIfRegistered<T>` extension methods
- Generated extensions are available at compile time for all registered modules
- Eliminates runtime type errors from incorrect module type parameters

## Changes
- New `ModuleExtensionsGenerator.cs` in `ModularPipelines.SourceGenerator`
- New `ModuleClassInfo.cs` record for tracking module information
- Generates extension methods in the user's namespace for easy discoverability

## Test Plan
- [ ] Verify generator compiles without errors
- [ ] Verify generated extensions appear in IntelliSense
- [ ] Verify calling GetModule<T> with registered type compiles
- [ ] Verify calling GetModule<T> with unregistered type fails at compile time

Closes #2006


thomhurst:
## Summary
This PR adds a source generator to create type-safe extension methods for module result retrieval.

## Critical Issues

### 1. Breaking Change: Build() to BuildCommandLine() Rename
Location: src/ModularPipelines.SourceGenerator/BuildMethodGenerator.cs:36

The rename from Build() to BuildCommandLine() is a breaking API change that will break all existing code using generated options classes. Every consumer using generated options classes (Git, Docker, DotNet, etc.) will break on upgrade.

Recommendation: Either keep Build() as the method name, OR clearly document as a breaking change with migration guidance.

### 2. Hardcoded Build Path in NuGet Package Target
Location: src/ModularPipelines/ModularPipelines.csproj:65

The hardcoded relative path assumes specific directory structure, won't work correctly if projects are built independently, and build order dependency is not enforced.

Recommendation: Use @(ReferencePath) or add explicit dependency to ensure correct build order.

## Suggestions

### 3. XML Documentation Escaping
Location: ModuleExtensionsGenerator.cs:173,182
Consider using System.Security.SecurityElement.Escape() for complete XML escaping.

### 4. Method Naming Edge Case
Location: ModuleExtensionsGenerator.cs:212-218
StripModuleSuffix() doesn't handle edge cases like module named just "Module" (returns empty string).

### 5. Duplicate Module Name Handling
Location: ModuleExtensionsGenerator.cs:168
If two modules in different namespaces have the same name, generated method will be ambiguous.

## Verdict
REQUEST CHANGES - Critical issues found: Breaking API change and NuGet packaging target issues

thomhurst:
## Addressed Review Feedback

### 1. Build() to BuildCommandLine() Rename (ADDRESSED)

Added documentation to `BuildMethodGenerator.cs` explaining:
- **Why renamed**: The `Build()` method conflicts with properties named "Build" (e.g., `AptGetOptions` has a `Build` property for the `--build` flag). In C#, a property and parameterless method cannot share the same name.
- **Non-breaking**: This method is internal infrastructure called by `ICommandLineBuilder`. Users don't call it directly on options instances - they pass options to command methods like `context.Git().Checkout(options)`.

### 2. StripModuleSuffix Edge Case (ADDRESSED)

Added documentation clarifying that the edge case for a class named exactly "Module" is already handled:
- The condition `className.Length > suffix.Length` ensures "Module" won't be stripped to empty string
- Added explicit example in the doc comments

### 3. NuGet Packaging

The current approach works because:
- Line 57 in `ModularPipelines.csproj` has `ProjectReference` with `ReferenceOutputAssembly="false" OutputItemType="Analyzer"` which establishes build dependency
- The target `_AddSourceGeneratorToPackage` runs after the source generator is built

The alternative pattern (separate `.Package` project) could be a follow-up improvement, but the current implementation is functional.

### 4. Duplicate Module Names

If two modules in different namespaces have the same class name, the generator uses `moduleGroup.First()`. This is consistent with how `GetModule<T>()` works - it requires unique module types. Users with duplicate names should use fully qualified types. A future enhancement could add a diagnostic warning.

PR #2084, opened 2026-01-16
feat: expand console width for CI environments

### LINKED ISSUE #2083: Spectre console width
Can we expand the output to the entire available width without wrapping? Currently, it is not very convenient to read logs in CI. All system logs outside ModularPipeline are written to the entire width, while ModularPipeline is limited to the default 80 characters

Image

google:
```text
Spectre.Console automatically detects your terminal's width but defaults to 80 columns
if it can't detect it (like when redirecting output), causing line wrapping;
you can control this by using methods like Table().Expand() to fill available space, 
setting fixed widths for columns (Table().Width()), or manually overriding the width in your code when needed, 
especially for redirected output where you might set it to a very large number or disable wrapping,
as discussed on the Spectre.Console GitHub issues https://github.com/spectreconsole/spectre.console/issues/1782
```

refs:
https://github.com/spectreconsole/spectre.console/issues/1867
https://github.com/spectreconsole/spectre.console/issues/1782

thomhurst:
## Summary

**Console Width Improvements:**
- Adds `ConsoleWidth` option to `PipelineOptions` for explicit width control
- Auto-detects known CI environments and expands width to 160 characters (from the default 80)
- Supports all 8 known CI systems: GitHub Actions, Azure Pipelines, TeamCity, GitLab, Jenkins, Bitbucket, Travis CI, and AppVeyor

**Duplicate Timestamp Fix:**
- Removes duplicate timestamps from log output
- The Spectre.Console logger was adding timestamps, but command logging already includes precise timestamps with milliseconds `[HH:mm:ss.fff]`
- Configures output templates for all log levels to exclude the `{DateTime}` placeholder

## Problem

1. **Console width**: Spectre.Console defaults to 80 characters when it cannot detect terminal width, which is common in CI environments where stdout is redirected. This causes tables and formatted output to wrap unnecessarily.

2. **Duplicate timestamps**: Log lines showed timestamps twice - once from the Spectre logger and once from command logging:
   ```
   11:54:48 Info] [11:54:48.360] Command:
   ```

### Console Width
When a known CI environment is detected via `IBuildSystemDetector`, the console width is automatically expanded to 160 characters. Users can override:

```csharp
await PipelineHostBuilder.Create()
    .ConfigurePipelineOptions(options => options.ConsoleWidth = 200)
    .ExecutePipelineAsync();
```

### Timestamps
Configure Spectre logger output templates to only show log level (e.g., `[Info]`) without timestamps, since command logging already provides precise timestamps.

## Test plan

- [x] Verify build passes
- [ ] Test locally with `ShowProgressInConsole = true` to ensure width still auto-detects correctly
- [ ] Test in GitHub Actions to confirm expanded width and single timestamps

Fixes #2083

### DIFF
--- a/src/ModularPipelines/DependencyInjection/DependencyInjectionSetup.cs
+++ b/src/ModularPipelines/DependencyInjection/DependencyInjectionSetup.cs
@@ -78,6 +78,34 @@ internal static class DependencyInjectionSetup
 
                 builder.AddSpectreConsole(config =>
                 {
+                    // Configure output templates without timestamps
+                    // Command logging already includes precise timestamps ([HH:mm:ss.fff])
+                    // so we don't need them from the logger as well
+                    config.ConfigureProfile(LogLevel.Trace, profile =>
+                    {
+                        profile.OutputTemplate = "[grey][[Trace]][/] {Message}{NewLine}{Exception}";
+                    });
+                    config.ConfigureProfile(LogLevel.Debug, profile =>
+                    {
+                        profile.OutputTemplate = "[grey][[Debug]][/] {Message}{NewLine}{Exception}";
+                    });
+                    config.ConfigureProfile(LogLevel.Information, profile =>
+                    {
+                        profile.OutputTemplate = "[deepskyblue1][[Info]][/] {Message}{NewLine}{Exception}";
+                    });
+                    config.ConfigureProfile(LogLevel.Warning, profile =>
+                    {
+                        profile.OutputTemplate = "[yellow][[Warn]][/] {Message}{NewLine}{Exception}";
+                    });
+                    config.ConfigureProfile(LogLevel.Error, profile =>
+                    {
+                        profile.OutputTemplate = "[red][[Error]][/] {Message}{NewLine}{Exception}";
+                    });
+                    config.ConfigureProfile(LogLevel.Critical, profile =>
+                    {
+                        profile.OutputTemplate = "[white on red][[Critical]][/] {Message}{NewLine}{Exception}";
+                    });
+
                     config.ConfigureProfiles(profile =>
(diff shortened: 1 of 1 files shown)


thomhurst:
## Summary
Adds configurable console width support with automatic expansion to 160 characters in CI environments and removes duplicate timestamps from log output.

## Critical Issues
None found ✅

## Suggestions

### 1. Missing null check in ConfigureConsoleWidth (src/ModularPipelines/Console/ConsoleCoordinator.cs:354)
The `_buildSystemDetector` field is injected via constructor but the `ConfigureConsoleWidth()` method uses it without a null check. While dependency injection should ensure it's not null, consider defensive programming since this is called during `Install()` which has error handling.

However, looking at the constructor, the field is required and will never be null in practice, so this is a minor observation rather than a blocking issue.

### 2. Consider property validation for ConsoleWidth (src/ModularPipelines/Options/PipelineOptions.cs:194)
The `ConsoleWidth` property accepts any `int?` value. Consider adding validation to ensure reasonable bounds (e.g., minimum width of 40, maximum of 500) to prevent unexpected behavior. This could be done via:
- A property setter with validation
- A separate validator class (similar to `ConcurrencyOptionsValidator` and `HttpResilienceOptionsValidator` already in the codebase)

This is optional but would improve robustness.

## Verdict

The implementation is solid:
- Correctly injects `IBuildSystemDetector` via DI
- Provides sensible defaults (160 chars for CI, auto-detect for local)
- Allows user override via `PipelineOptions.ConsoleWidth`
- Log template changes cleanly remove duplicate timestamps
- Good documentation in XML comments

The suggestions above are optional improvements for defensive programming and validation, but don't block the PR.

PR #2158, opened 2026-01-17
feat(build): add VersionTag category and workflow for automatic tagging

thomhurst:
## Summary
- Add `[ModuleCategory("VersionTag")]` attribute to `NugetVersionGeneratorModule` and `PushVersionTagModule`
- Add support for `Pipeline__RunCategories` environment variable in Program.cs to run only specific module categories
- Create new `version-tag.yml` workflow that triggers on push to main and runs only the VersionTag category modules

## Test plan
- [ ] Verify the workflow triggers on push to main
- [ ] Verify only VersionTag modules run when `Pipeline__RunCategories=VersionTag` is set
- [ ] Verify normal pipeline execution is unaffected when env var is not set

### DIFF
--- a/.github/workflows/version-tag.yml
+++ b/.github/workflows/version-tag.yml
@@ -10,8 +10,6 @@ permissions:
 jobs:
   version-tag:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
 
     steps:
       - uses: actions/checkout@v6


thomhurst:
## Summary
Adds module categories and a dedicated workflow to automatically create and push version tags after commits to main.

## Critical Issues

**1. Missing Git Authentication Setup** 🚨

The `PushVersionTagModule` will fail to push tags because git authentication is not configured.

**Problem**: In `PushVersionTagModule.cs:36-47`, the module calls `git push --tags` without setting up authentication. The workflow sets `persist-credentials: false` (.github/workflows/version-tag.yml:14), so the default GitHub token won't work.

**Evidence**: Other modules that push to git (e.g., `FormatMarkdownModule.cs:93-97`) use `GitHelpers.SetUserCommitInformation()` and `GitHelpers.CheckoutBranch()` to configure the remote URL with the access token before pushing.

**Fix needed**: Before creating and pushing the tag, the module must:
1. Call `GitHelpers.SetUserCommitInformation(context, cancellationToken)` to set git user
2. Update the remote URL with the access token using `git remote set-url origin https://x-access-token:{token}@github.com/...`

See `GitHelpers.cs:47-65` and `FormatMarkdownModule.cs:93-98` for reference implementations.

## Suggestions

**1. Consider Adding Skip Logic for Missing Token**

The `PushVersionTagModule` should skip execution if `GitHub__StandardToken` is not available, similar to `FormatMarkdownModule.cs:40-43`. This would make the module more resilient in development environments.

Suggestion: Add a `ShouldSkip()` override or use the `[SkipIfNoStandardGitHubToken]` attribute.

**2. Workflow Naming Consistency**

The workflow file uses "Version Tag" but the category name is "VersionTag" (no space). Consider whether they should match for clarity.

## Previous Review Status
No previous comments found.

## Verdict
⚠️ **REQUEST CHANGES** - Critical authentication issue will cause the workflow to fail

PR #2200, opened 2026-01-18
[Automated] Update buildah CLI Options

thomhurst:
## Summary
This PR contains automatically generated updates to **buildah** CLI options classes.

## Changes

## Verification

---


thomhurst:
## Summary
Automated generator update to buildah CLI options classes, removing `#nullable enable` directives and changing nullable reference type annotations in method signatures.

## Critical Issues
⚠️ **Breaking Change**: The PR changes method signatures from nullable parameters (`BuildahTagOptions? options = default`) to non-nullable parameters (`BuildahTagOptions options = default`). This is a **breaking API change** that could cause compilation errors for consumers explicitly passing `null`:

```csharp
// This would break after the change:
await buildah.Tag(options: null);  // CS1503: Argument type 'null' cannot convert to 'BuildahTagOptions'
```

**Affected methods**: All 24 method signatures in `IBuildah` interface changed from `?options = default, ?executionOptions = null` to `options = default, executionOptions = null`.

**Impact**: While `default` works for both nullable and non-nullable value types/classes, consumers explicitly using `null` will break.

## Suggestions
1. **Version bump consideration**: Since this is a breaking API change, this should trigger at least a minor version bump (or major if following strict semver).

2. **Generator consistency**: The `[GeneratedCode]` attribute now has an empty version string (`""` instead of `"2.42.227-alpha.0.981+..."`). Consider whether the generator version should be tracked for diagnostics.

3. **`#nullable enable` removal**: Removing explicit nullable annotations while simultaneously changing signatures from nullable to non-nullable seems inconsistent. Without `#nullable enable`, the non-nullable annotations have no effect in projects not using nullable reference types globally.

## Previous Review Status
No previous comments accessible (API scope limitation).

## Verdict
⚠️ **REQUEST CHANGES** (if not yet merged) - Breaking API change requires version planning

**Note**: PR appears to be already merged. If so, recommend:
- Publishing this as a new minor/major version
- Documenting the breaking change in release notes
- Considering whether to revert and re-evaluate the nullable annotation strategy

PR #2202, opened 2026-01-18
[Automated] Update pulumi CLI Options

thomhurst:
## Summary
This PR contains automatically generated updates to **pulumi** CLI options classes.

## Changes

## Verification

---


thomhurst:
## Summary
Updates auto-generated Pulumi CLI options classes to reflect latest CLI documentation.

## Critical Issues

**Nullable Reference Type Inconsistency** ❌

The generator removed `#nullable enable` directives from all generated files and removed nullable annotations (`?`) from method parameters. This creates a critical inconsistency:

1. **Project-level nullable is enabled**: `Directory.Build.props` sets `<Nullable>enable</Nullable>` globally
2. **Generated files disable nullable**: Removing `#nullable enable` from generated files disables nullable checking in those files
3. **Uninitialized fields**: Fields like `private PulumiConfigEnv _env;` are no longer marked nullable but aren't initialized in constructors

**Specific problems:**

```csharp
// In PulumiConfig.Generated.cs
private PulumiConfigEnv _env;  // No nullable annotation, no initialization

public PulumiConfigEnv Env => _env ??= new PulumiConfigEnv(_command);
```

Without `#nullable enable`, this field is implicitly nullable but:
- Won't generate CS8618 warnings about uninitialized non-nullable field
- Silently breaks the project's nullable reference type contract
- Could cause runtime null reference exceptions if accessed incorrectly

**Method signature changes:**

```csharp
// Before
Task<CommandResult> Cancel(PulumiCancelOptions? options = default, CommandExecutionOptions? executionOptions = null, ...)

// After  
Task<CommandResult> Cancel(PulumiCancelOptions options = default, CommandExecutionOptions executionOptions = null, ...)
```

Removing `?` from parameters with `= default` means they're non-nullable but default to null, which violates nullable semantics when nullable is enabled.

**Impact**: This compiles without warnings only because `#nullable enable` was removed. With nullable enabled (as per project settings), this code should generate CS8618 and CS8625 warnings.

**Recommendation**: The generator should either:
- Keep `#nullable enable` and use proper nullable annotations (`?` for nullable, `!` for null-forgiving, or initialization)
- Use `#nullable disable` explicitly to document the intent
- Initialize non-nullable fields properly in constructors

## Suggestions

**Empty GeneratedCode version string**: The `[GeneratedCode]` attribute version changed from `"2.42.227-alpha.0.981+55c59948e83059a5184db05c992047e177e72c4b"` to `""`. Consider including the generator version for traceability.

## Verdict
⚠️ **REQUEST CHANGES** - Critical nullable reference type inconsistency

PR #2210, opened 2026-01-18
[Automated] Update winget CLI Options

thomhurst:
## Summary
This PR contains automatically generated updates to **winget** CLI options classes.

## Changes

## Verification

---


thomhurst:
## Summary
Updates winget CLI options classes with auto-generated code changes from the latest CLI help output.

## Critical Issues
None found ✅

## Suggestions

**Nullability Contract Change**
The parameter signature change from nullable to non-nullable reference types (e.g., `WingetConfigureOptions? options` → `WingetConfigureOptions options`) with `= default` is semantically confusing. When a reference type parameter has `= default`, the actual default value is `null`, but declaring it as non-nullable suggests it should never be null.

Consider one of these approaches:
1. Keep nullable types: `WingetConfigureOptions? options = null`
2. Remove default and make required: `WingetConfigureOptions options` (but this breaks API compatibility)
3. Document that `default` means null for these parameters

This is a generated code pattern, so if this is intentional, consider updating the generator to make the intent clearer.

**GeneratedCode Version String**
The `GeneratedCode` attribute version changed from a valid version string to an empty string. While this doesn't affect runtime behavior, it removes useful metadata about which generator version produced the code. Consider preserving the version information.

## Verdict

The changes are consistent auto-generated updates. The nullability pattern and missing version are design decisions in the generator rather than defects in this PR.

PR #2227, opened 2026-01-18
feat: defer module output until pipeline end when live progress enabled

thomhurst:
## Summary

- When live progress is enabled, module output is now deferred until the pipeline ends instead of flushing immediately after each module completes
- Output is grouped by module and displayed in completion order after progress stops
- Prevents console flickering caused by repeated progress pause/resume cycles during module output flushing

## Changes

- Added `SetProgressActive`, `OnModuleCompletedAsync`, and `FlushDeferredAsync` methods to `IOutputCoordinator`
- `OutputCoordinator` stores deferred buffers with timestamps and flushes them in order at pipeline end
- `ModuleLogger` now calls `OnModuleCompletedAsync` which decides whether to defer or flush immediately
- `PipelineOutputCoordinator` calls `FlushDeferredAsync` after progress ends, before results are printed
- Added `IAsyncDisposable` safety net to ensure deferred output is never lost

## Test Plan

- [x] Unit tests for deferred flush behavior pass
- [x] Full test suite passes (900+ tests)
- [x] Build succeeds with no errors

### DIFF
--- a/src/ModularPipelines/Console/OutputCoordinator.cs
+++ b/src/ModularPipelines/Console/OutputCoordinator.cs
@@ -8,7 +8,7 @@ namespace ModularPipelines.Console;
 /// Coordinates immediate flushing of module output with FIFO ordering and synchronization.
 /// </summary>
 [ExcludeFromCodeCoverage]
-internal sealed class OutputCoordinator : IOutputCoordinator, IAsyncDisposable
+internal sealed class OutputCoordinator : IOutputCoordinator
 {
     private readonly IBuildSystemFormatterProvider _formatterProvider;
     private readonly ILoggerFactory _loggerFactory;
@@ -60,17 +60,31 @@ internal sealed class OutputCoordinator : IOutputCoordinator, IAsyncDisposable
     /// <inheritdoc />
     public void SetProgressActive(bool isActive)
     {
-        if (isActive)
+        lock (_deferredLock)
         {
-            // Starting a new progress session - check for stale deferred outputs
-            // from a previous run that crashed before FlushDeferredAsync was called
-            lock (_deferredLock)
+            if (isActive)
             {
+                // Starting a new progress session - check for stale deferred outputs
+                // from a previous run that crashed before FlushDeferredAsync was called
                 if (_deferredOutputs.Count > 0)
                 {
                     _logger.LogWarning(
                         "Found {Count} stale deferred outputs from a previous pipeline run. " +
                         "This indicates FlushDeferredAsync was not called. Clearing to prevent memory leak.",
+                        _deferredOutputs.Count);
+                    _deferredOutputs.Clear();
+                }
+            }
+            else
+            {
+                // Progress ending - clean up any remaining deferred outputs that weren't flushed
+                // This handles cancellation scenarios where FlushDeferredAsync wasn't called
+                if (_deferredOutputs.Count > 0)
+                {
+                    _logger.LogWarning(
+                        "Progress ended with {Count} unflushed deferred outputs. " +
+                        "This indicates FlushDeferredAsync was not called (possibly due to cancellation). " +
+                        "Clearing to prevent memory leak.",
                         _deferredOutputs.Count);
                     _deferredOutputs.Clear();
                 }
@@ -267,34 +281,6 @@ internal sealed class OutputCoordinator : IOutputCoordinator, IAsyncDisposable
         public PendingFlush(IModuleOutputBuffer buffer)
         {
             Buffer = buffer;
-        }
-    }
-
-    /// <summary>
-    /// Safety net: flushes any remaining deferred output on disposal.
-    /// </summary>
-    /// <remarks>
-    /// <para>
-    /// This is a defensive measure - the primary flush happens in
-    /// <see cref="Engine.Executors.PipelineOutputCoordinator.PipelineOutputScope.DisposeAsync"/>
-    /// which explicitly calls <see cref="FlushDeferredAsync"/> after progress ends.
-    /// </para>
-    /// <para>
-    /// This safety net handles edge cases where the normal flush sequence is interrupted
-    /// (e.g., due to exceptions or abnormal termination). Note that this requires the
-    /// DI container to be disposed via <c>DisposeAsync()</c> - synchronous disposal
-    /// may not trigger this method.
-    /// </para>
-    /// </remarks>
-    public async ValueTask DisposeAsync()
-    {
-        try
-        {
-            await FlushDeferredAsync(CancellationToken.None).ConfigureAwait(false);
-        }
-        catch (Exception ex)
-        {
-            _logger.LogWarning(ex, "Failed to flush deferred output during disposal");
         }
     }
 }


thomhurst:
## Summary
Defers module output until pipeline end when live progress is enabled to prevent console flickering from pause/resume cycles.

## Critical Issues

### 1. Race Condition in Progress Active State (ConsoleCoordinator.cs:232-233)
```csharp
// Wire up the progress controller for output coordination
_outputCoordinator.SetProgressController(session);
_outputCoordinator.SetProgressActive(true);  // Added in PR
```

**Issue**: `SetProgressActive(true)` is called **outside** the `_phaseLock` (lock released at line 222), while `_isProgressActive = true` is set **inside** the lock (line 221). This creates a race condition where:
- ConsoleCoordinator thinks progress is active
- OutputCoordinator thinks progress is NOT active yet
- Module output could be flushed immediately instead of deferred

**Fix**: Move `_outputCoordinator.SetProgressActive(true)` inside the lock before line 222:
```csharp
lock (_phaseLock)
{
    if (_isProgressActive) { ... }
    _isProgressActive = true;
    _outputCoordinator.SetProgressActive(true);  // Move inside lock
}
```

### 2. Missing SetProgressActive(false) in ConsoleCoordinator.cs:249

The `EndProgressPhase()` method sets progress to inactive, but the code shows:
```csharp
_outputCoordinator.SetProgressController(NoOpProgressController.Instance);
_outputCoordinator.SetProgressActive(false);
```

However, this is **inside** the `_phaseLock`, which is correct. But there's an asymmetry issue: if `BeginProgressAsync` returns a NoOpProgressSession when progress is disabled (line 210), the OutputCoordinator's progress state is never set. This means if progress was previously active, then disabled, then re-enabled, the state could be stale.

**Recommendation**: Also call `_outputCoordinator.SetProgressActive(false)` when creating NoOpProgressSession (around line 210).

### 3. Memory Leak - Unbounded Deferred Output Growth (OutputCoordinator.cs:76-80)

```csharp
private readonly List<DeferredModuleOutput> _deferredOutputs = new();
```

**Issue**: If a pipeline runs for a very long time with many modules and progress enabled, `_deferredOutputs` will grow without bound until `FlushDeferredAsync()` is called at the very end. For pipelines with thousands of modules, this could consume significant memory.

**Severity**: Medium - unlikely in most scenarios but possible in very large pipelines.

**Recommendation**: Consider a maximum size or periodic flushing strategy for extremely long-running pipelines.

### 4. PipelineOutputCoordinator Missing Parameter in Constructor (PipelineOutputCoordinator.cs:30-42)

The PR diff shows `IOutputCoordinator` is added as a parameter, but the original file (read at line 30-42) **doesn't have it**. Looking at the diff again:

```diff
+    private readonly IOutputCoordinator _outputCoordinator;
+        IOutputCoordinator outputCoordinator)
+        _outputCoordinator = outputCoordinator;
```

This means the PR correctly adds the dependency. However, I notice the original file I read doesn't have this - **this is expected since I read the current main branch, not the PR branch**. This is correct in the PR.

### 5. Timeout Message Change is Misleading (ModuleLogger.cs:173-183)

The diff changes the timeout warning message from:
```csharp
-"Module output flush timed out after 30 seconds for {ModuleType}. Some output may be lost."
+"Module output handling timed out after 30 seconds for {ModuleType}."
```

**Issue**: The new message removes "Some output may be lost", which is **critical information** for debugging. When output is deferred and the timeout occurs, output **will** be lost since it won't be flushed.

**Recommendation**: Keep the "Some output may be lost" warning or change to "Output may be deferred until pipeline end."

## Suggestions

### 1. Test Coverage for Race Conditions
The new tests (OutputCoordinatorDeferredFlushTests.cs) only test happy paths. Consider adding tests for:
- Concurrent module completion during progress state changes
- FlushDeferredAsync called while modules are still completing
- Disposal race conditions

### 2. Documentation Clarity
The comment in OutputCoordinator.cs:76 says "defer output until pipeline end", but this assumes progress is enabled. Consider clarifying: "defer output until progress ends (if progress is active)"

## Previous Review Status
No previous review comments found (permission issues prevented reading comments API).

## Verdict
⚠️ **REQUEST CHANGES** - Critical race condition (#1) must be fixed before merge. Issues #2 and #5 should also be addressed.

thomhurst:
## Summary
Defers module output until pipeline end when live progress is enabled to prevent console flickering from repeated pause/resume cycles.

## Critical Issues

### 1. IAsyncDisposable Won't Be Called - Safety Net is Ineffective

`OutputCoordinator` implements `IAsyncDisposable` (OutputCoordinator.cs:11) as a safety net to flush deferred output on disposal. However, `OutputCoordinator` is registered as a **singleton service**, and the .NET DI container does **not** automatically dispose `IAsyncDisposable` services.

**Impact**: The safety net `DisposeAsync()` method at line 290 will **never be called**, making it dead code. This creates a false sense of safety - if `FlushDeferredAsync()` is not called in `PipelineOutputCoordinator.DisposeAsync()`, deferred output will be permanently lost.

**Recommendation**: Remove the `IAsyncDisposable` implementation entirely. The extensive documentation (lines 272-289) acknowledges this is a "defensive measure" but also states the primary flush happens in `PipelineOutputCoordinator`. Since the disposal path never executes, keeping it is misleading.

---

### 2. Memory Leak on Pipeline Cancellation

When `_isProgressActive` is true, `OnModuleCompletedAsync()` (line 84) adds buffers to `_deferredOutputs`. These are only cleared in two places:
1. `FlushDeferredAsync()` - called at pipeline end
2. `SetProgressActive(true)` - clears stale outputs from previous runs (line 70)

**Issue**: If a pipeline is **cancelled mid-execution** (before reaching `PipelineOutputCoordinator.DisposeAsync()`), the deferred outputs remain in memory for the lifetime of the singleton `OutputCoordinator`.

**Impact**: In long-running applications that execute multiple pipelines (e.g., a web service running CI/CD pipelines), this accumulates unbounded memory.

**Example scenario**:
1. Pipeline starts with progress enabled
2. 100 modules complete → 100 buffers added to `_deferredOutputs`
3. User cancels pipeline (Ctrl+C)
4. `PipelineOutputCoordinator.DisposeAsync()` might not be called
5. Next pipeline starts → previous 100 buffers still in memory
6. Repeat over days → memory leak

**Recommendation**: Clear `_deferredOutputs` in `SetProgressActive(false)` at line 258 (in ConsoleCoordinator.EndProgressPhase). This ensures cleanup happens whenever progress ends, regardless of cancellation:

```csharp
public void SetProgressActive(bool isActive)
{
    if (!isActive)
    {
        // Progress ending - clean up any remaining deferred outputs
        lock (_deferredLock)
        {
            if (_deferredOutputs.Count > 0)
            {
                _logger.LogWarning(
                    "Progress ended with {Count} unflushed deferred outputs. " +
                    "This indicates FlushDeferredAsync was not called. Clearing to prevent memory leak.",
                    _deferredOutputs.Count);
                _deferredOutputs.Clear();
            }
        }
    }
    else
    {
        // ... existing stale output check ...
    }
    
    _isProgressActive = isActive;
}
```

## Suggestions

### Test Coverage for Error Scenarios

The new test file `OutputCoordinatorDeferredFlushTests.cs` only tests happy paths with progress disabled. Consider adding tests for:
- Pipeline cancellation before flush
- Multiple concurrent module completions (race condition stress test)  
- Verifying deferred outputs are cleared when progress ends without flush

## Previous Review Status

**First Review Issues:**
- ✅ Race condition in `SetProgressActive`: **FIXED** - now inside lock
- ✅ Missing `SetProgressActive(false)` for NoOpProgressSession: **FIXED**
- ⚠️ Unbounded memory growth: **PARTIALLY ADDRESSED** - see Critical Issue #2
- ✅ Timeout message: **FIXED** - retains "Some output may be lost"

**Second Review Issues:**  
- ❌ IAsyncDisposable won't be called: **NOT FIXED** - see Critical Issue #1
- ⚠️ Memory leak on cancellation: **PARTIALLY ADDRESSED** - see Critical Issue #2
- ✅ Lock contention: **FIXED** - separate `_deferredLock` added

## Verdict
⚠️ **REQUEST CHANGES** - Two critical issues must be addressed:
1. Remove ineffective `IAsyncDisposable` implementation (misleading dead code)
2. Clear `_deferredOutputs` when progress ends to prevent memory leaks on cancellation

Issue #3015, opened 2026-07-21
generate-cli-options workflow has failed every week since 2026-01-18 — tracking issue for the 5 root-cause classes


thomhurst:
## Summary

Every scheduled run of `generate-cli-options.yml` has failed since the last success on **2026-01-18** — 35 failures and 4 cancellations in the last 40 runs, six months of no generated-options updates reaching `main`. Latest analyzed run: [29671358939](https://github.com/thomhurst/ModularPipelines/actions/runs/29671358939) (2026-07-19), 18 of the matrix jobs failed.

## Root causes (one issue per class)

| # | Class | Tools affected |
|---|-------|----------------|
| #3010 | Scrapers return zero commands (install failures + invocation/parse bugs) | aws, gh, snyk, yarn, cosign, eksctl, trivy, liquibase |
| #3011 | Generated service instantiates abstract options class (CS0144, single-command tools) | ansible, hadolint, jq, shellcheck, sonar-scanner, mvn, gradle |
| #3012 | Duplicate Compile items / AppSet-Appset casing (NETSDK1022) | argocd |
| #3013 | Partial generation — extensions emitted without service interface (CS0246) | kubectl |
| #3014 | Manual builder drift vs regenerated options (CS0117) | dotnet |

Passing jobs in the same run: helm, docker, git, terraform, vault, yq, node, choco, winget, and others — the pipeline itself works when scraping and codegen succeed.

## Consequences

- The 12 empty NuGet stub packages (#2998–#3009) are a direct symptom: classes #3010/#3011/#3012 are exactly why their generation never landed.
- Implemented packages (DotNet, Kubernetes, AWS wrappers, GitHub, Yarn) have received no options updates in six months.
- Nobody was alerted: the workflow fails weekly with no notification and no badge.

## Scope

1. Fix the five classes above (tracked individually).
2. Add failure alerting: open/update a single tracking issue automatically on scheduled-run failure, or add a workflow status badge to the README.
3. Related hardening: #2992 (API-surface gate + fail on zero output for matrix tools).

Issue #3013, opened 2026-07-21
OptionsGenerator: kubectl job emits KustomizeExtensions without IKustomize service (CS0246) — partial generation is not transactional


thomhurst:
Part of the `generate-cli-options.yml` failure investigation (every weekly run has failed since 2026-01-18). Evidence from run [29671358939](https://github.com/thomhurst/ModularPipelines/actions/runs/29671358939) (2026-07-19).

## Failure

The `kubectl` job regenerates `ModularPipelines.Kubernetes` and the compile check fails:

```
Extensions/KustomizeExtensions.Generated.cs(49,19): error CS0246: The type or namespace name 'IKustomize' could not be found
```

## Analysis

Partial/inconsistent generation: the extensions file referencing the Kustomize service was emitted (or left in place from a previous run), but `Services/IKustomize.Generated.cs` was not generated in this run — likely the kustomize sub-tool scrape failed silently while the extensions emitter still ran, or stale generated files from the checkout are mixed with a partial regeneration that deleted/skipped the service.

## Scope

1. Make generation transactional per tool: clean the package's `*.Generated.cs` output and emit the full consistent set, or fail before writing anything if any sub-tool scrape fails.
2. Ensure a scrape failure for a sub-tool (kustomize) is a loud error, not a silent skip.

thomhurst:
Resolved by #3016 (merged in commit 8604c5a5a). Verified on current main: Generator_Failure_Leaves_Existing_Output_Untouched, Zero_Commands_For_Cli_Only_Tool_Is_An_Error_And_Writes_Nothing, and Unavailable_Cli_Only_Tool_Is_An_Error_Not_A_Silent_Skip all pass; the OptionsGenerator test project builds successfully.

Issue #3012, opened 2026-07-21
OptionsGenerator: argocd generation produces duplicate Compile items (NETSDK1022), AppSet/Appset casing inconsistency


thomhurst:
Part of the `generate-cli-options.yml` failure investigation (every weekly run has failed since 2026-01-18). Evidence from run [29671358939](https://github.com/thomhurst/ModularPipelines/actions/runs/29671358939) (2026-07-19).

## Failure

The `argocd` job scrapes and generates successfully, then the compile check fails:

```
error NETSDK1022: Duplicate 'Compile' items were included. ... The duplicate items were:
'Enums/ArgoCdAppSetLogformat.Generated.cs'; 'Enums/ArgoCdAppsetLoglevel.Generated.cs'; 'Options/ArgoCdAppsetOptions.Generated.cs'
```

## Analysis

NETSDK1022 means the same files are included both by the SDK's implicit `Compile` glob and by explicit `Compile` items. `src/ModularPipelines.ArgoCd.csproj` is a bare stub with no explicit items, so the explicit includes are presumably being written during generation (generator emitting csproj item groups, or files emitted twice under case-variant names — note `AppSet` vs `Appset` casing inconsistency in the duplicate list, suggesting the command-name casing normalizer produces two spellings of the same command group).

## Scope

1. Reproduce locally with the ArgoCd scraper and inspect what the generator writes besides `.cs` files.
2. Fix the duplicate inclusion (and normalize `AppSet`/`Appset` casing while in there — case-variant filenames will also collide on case-insensitive filesystems).

Blocks the empty-stub package #2999 (ArgoCd).

Issue #3010, opened 2026-07-21
OptionsGenerator CI: scrapers return zero commands for 8 tools (aws, gh, snyk, yarn, cosign, eksctl, trivy, liquibase)


thomhurst:
Part of the `generate-cli-options.yml` failure investigation (every weekly run has failed since 2026-01-18). Evidence from run [29671358939](https://github.com/thomhurst/ModularPipelines/actions/runs/29671358939) (2026-07-19).

## Failure

Eight matrix jobs fail with:

```
System.InvalidOperationException: No commands found for <tool>. This is a CLI-only tool with no HTML fallback. Ensure the CLI tool is installed and accessible.
```

after logging `Finished scraping <tool>. Total commands: 0`.

Affected tools: **aws, gh, snyk, yarn, cosign, eksctl, trivy, liquibase**

## Per-tool evidence (causes differ within the class)

- **liquibase** — the install step itself fails: `Error: Could not find liquibase executable`. CLI never installed.
- **aws** — the AWS CLI errors when invoked by the scraper: `aws: [ERROR]: list index out of range` (twice), then 0 commands. Likely the scraper invokes a help form awscli v2 doesn't support.
- **gh** — GitHub CLI is preinstalled on `ubuntu-latest`, yet the scraper still finds 0 commands — scraper invocation/parse bug rather than a missing binary.
- **snyk, yarn, cosign, eksctl, trivy** — same 0-commands symptom; needs triage per tool (install step vs. scraper parse). Note argocd (also Cobra-based) scrapes successfully, so this is not a blanket Cobra-scraper break.

## Scope

1. For each tool, confirm whether the binary is installed and on PATH in the job (add a `<tool> --version` step to make this explicit in logs).
2. Fix install steps (liquibase at minimum) and scraper invocations (aws, gh).
3. Make the scrapers log the exact command they executed and its stderr on failure — the current "0 commands" summary hides the root cause.

Fixing this class unblocks the empty-stub packages #3000 (Cosign), #3001 (Eksctl), #3005 (Liquibase), #3007 (Snyk), #3009 (Trivy).

PR #3016, opened 2026-07-21
fix(options-generator): harden code generation against silent failures

### LINKED ISSUE #3011: OptionsGenerator: single-command tools generate 'new AbstractOptions()' — CS0144 blocks 7 packages (ansible, hadolint, jq, shellcheck, sonar-scanner, mvn, gradle)
Part of the `generate-cli-options.yml` failure investigation (every weekly run has failed since 2026-01-18). Evidence from run [29671358939](https://github.com/thomhurst/ModularPipelines/actions/runs/29671358939) (2026-07-19).

## Failure

Seven matrix jobs successfully scrape and generate code, but the post-generation compile check fails with the same error shape:

```
Services/<Tool>.Generated.cs(40,65): error CS0144: Cannot create an instance of the abstract type or interface '<Tool>Options'
```

Affected: **ansible** (`AnsibleOptions`), **hadolint** (`HadolintOptions`), **jq** (`JqOptions`), **shellcheck** (`ShellcheckOptions`), **sonar-scanner** (`SonarScannerOptions`), **mvn** (`MavenOptions`, in `ModularPipelines.Java`), **gradle** (`GradleOptions`, in `ModularPipelines.Java`).

## Analysis

All affected tools are single-root-command CLIs (no subcommand tree). The generator appears to emit the root options class as `abstract` (its usual pattern for shared base options of a command family) while the generated service at `Services/<Tool>.Generated.cs:40` instantiates it directly with `new <Tool>Options()`. For tools with subcommands, the instantiated classes are the concrete per-subcommand options, so the bug only bites single-command tools.

## Scope

In the generator's service/options emitters: when a tool has no subcommands (or a command has options and is itself invokable), emit the options class as concrete — or emit a concrete `<Tool>DefaultOptions` subclass and instantiate that.

This is the blocker behind the empty-stub packages #2998 (Ansible), #3002 (Hadolint), #3003 (Java — both Maven and Gradle), #3004 (Jq), #3006 (Shellcheck), #3008 (SonarScanner): generation works, the compile gate then rejects the output, so the PR is never created.

thomhurst:
## Summary

Fixes the generator-side root causes behind six months of weekly `generate-cli-options` failures (#3015), plus the hardening improvements from the generator review.

## Correctness fixes

- **Transactional generation** (#3013): all files are generated in memory and validated *before* anything on disk is touched. Previously `CleanupGeneratedFiles` ran before scraping, so a mid-scrape failure left packages half-deleted (the kubectl/kustomize `IKustomize` CS0246).
- **Duplicate output path detection** (#3011, #3012): generators producing the same path (case-insensitive) now fail loudly. Previously last-writer-won silently — which is exactly how single-command tools shipped an abstract options class where a concrete one was expected.
- **Single-command tool fix — CS0144** (#3011): root commands whose class name equals the abstract base (`AnsibleOptions : AnsibleOptions`) are renamed to `{Prefix}ExecuteOptions`. Unblocks ansible, hadolint, jq, shellcheck, sonar-scanner, mvn, gradle.
- **Case-insensitive enum dedup + one PascalCase implementation** (#3012): `AllEnums` deduplicates with `OrdinalIgnoreCase`, and the three separate casing implementations are consolidated into `ToPascalCase` — the source of `AppSet`/`Appset`-style case-variant file collisions.
- **Nullable emission**: optional parameters are now `{T}? options = null` / `CommandExecutionOptions? executionOptions = null` and lazy sub-domain fields are nullable. Generated files declare `#nullable enable`, so the previous output produced CS8625/CS8618 in every generated service (visible in shipped `Yq.Generated.cs`).
- **Shared collision filter**: `GetNonCollidingRootCommands` is now computed once and used by both the interface and implementation generators, so they can never disagree (CS0535 class of bugs).
- **CS0102 guard**: property-name dedup now spans required (primary-constructor) and optional members.
- **`ToDictionary` crash guard** in `SubDomainClassGenerator` for duplicate single-part commands.

## Diagnostics & safety

- **Zero-command diagnostics** (#3010): the error now distinguishes "CLI not on PATH" from "CLI available but help parsing produced nothing" (the old message said "ensure the CLI is installed" even when availability had just been verified). An unavailable CLI-only tool is now an error instead of a silent skip.
- **Hand-written files are never deleted**: removed the name-based deletion of `{Prefix}Extensions.cs`; only files carrying the `<auto-generated>` header marker are cleanup candidates.
- **No more guaranteed weekly diff**: `AssemblyInfo.Generated.cs` no longer embeds a generation timestamp, so no-change regenerations produce no diff and no PR.
- **Runner path leak** (#2991): `/home/runner`, `/Users/runner(admin)`, `C:\Users\runneradmin` in scraped defaults are normalized to `~` in generated XML docs.
- **`[GeneratedCode]` version** (#2997, partial): stamped with the generator assembly version instead of `""`.
- `Passphrase` added to the secret keyword list.

## Behavior changes to be aware of

- The next weekly generation run will re-emit all generated files once (nullable signatures, `[GeneratedCode]` version, AssemblyInfo shape). One-time churn.
- Method names for command parts containing `_` or spaces now PascalCase consistently (previously only `-` was split). No committed generated output is affected.
- CLI-only tools whose CLI is missing now fail the run instead of silently skipping.

## Testing

- `GeneratorHardeningTests` added covering normalization, collision detection, nullable emission, casing, enum dedup, the collision filter, path normalization, and the version stamp.
- Full generator suite: **323/323 pass**, solution builds with zero new warnings.

Closes #3011. Contributes to #3010, #3012, #3013, #3015, #2991, #2997.


thomhurst:
Addressed the review feedback in 6e143e8eb:

**Silent data loss on duplicates (greptile P1 x3, claude #1)** — all three duplicate classes now fail loudly instead of picking arbitrary survivors:
- `AllEnums` reverted to case-*sensitive* dedup. Dropping a case-variant silently left options referencing the dropped spelling (C# type lookup is case-sensitive), so the ignore-case dedup was masking the bug. Case variants now flow through to generated file paths, where `EnsureNoDuplicateFilePaths` fails the tool naming both spellings.
- Duplicate single-part root command definitions throw in `SubDomainClassGenerator` with the colliding command names, instead of `First()`-wins.
- Distinct commands normalizing to the same method name (`build-server` vs `build_server` → `BuildServer`) throw in `GetNonCollidingRootCommands` — this collision lives inside one file, so the path check can't see it.

**Orchestrator test coverage (claude #2)** — added `CodeGeneratorOrchestratorTests` with fake `ICliScraper`/`ICodeGenerator`: generator failure, duplicate paths, zero commands, and unavailable CLI all leave pre-existing output untouched and record errors; successful generation replaces old output. 331/331 tests pass.

**Non-atomic write window (greptile inline, claude #3)** — accepted as-is, doc comment softened to state the actual guarantee: validated-before-mutating, not crash-atomic. An I/O failure mid-write-loop is caught by the post-generation compile check before any PR is created; a temp-dir + rename swap can follow up if it ever bites in practice.

thomhurst:
Greptile's remaining P1 (non-transactional final writes) addressed in 25c2113c5 — went with a fix rather than the accepted-risk stance:

**Write-then-prune**: the delete-old-then-write order is inverted. All replacements are written first (overwriting in place where paths match), and only then are stale files — prefix-matched generated files *not* written this run — pruned. Nothing is ever deleted until its replacement exists on disk. A cancellation, permission error, or full disk mid-write now leaves the previous files intact (at worst alongside some new ones, which the compile check catches) instead of a half-deleted package. New orchestrator test pins the guarantee: a mid-loop write failure preserves pre-existing output because pruning never ran. 332/332 tests pass.

Note for the claude bot's re-review (17:18): it examined the intermediate commit `a6419cfb` — every item listed as "still open" (silent duplicate-command drop, AllEnums case-variant drop, missing orchestrator tests, doc overstatement) was already fixed in `6e143e8eb`, and the write-window item is now fixed by this commit.

PR #3019, opened 2026-07-21
Remove unused module state store

### LINKED ISSUE #2986: Remove unused ModuleStateStore immutable state machine (dead follow-up to #1997)
## Summary

The immutable state machine added for #1997 (`src/ModularPipelines/Engine/State/`) was merged on 2026-01-13 but never wired into the engine. The scheduler and runner still use the mutable `ModuleState` + `ReaderWriterLockSlim` path. Six months on, it is dead code.

## Evidence

- Referenced only by its own files, its DI registration (`DependencyInjectionSetup.cs`, ~line 327), and its unit tests — nothing in the scheduler/runner consumes it.
- Its XML docs claim "This class replaces the previous ReaderWriterLockSlim-based synchronization" — which never happened, misleading contributors.
- It duplicates completion ownership: `ModuleStateSnapshot.CompletionSource` vs the real TCS in `Module<T>` (`Modules/Module.cs`, ~line 52). Two sources of truth for completion is exactly the problem #1997 set out to fix.

## Scope

Delete:
- `src/ModularPipelines/Engine/State/` (5 files: `IModuleStateStore`, `ModuleStateStore`, `ModuleStateSnapshot`, `ModuleExecutionPhase`, `ModuleStateTransitions`)
- The DI registration in `DependencyInjectionSetup.cs`
- `test/ModularPipelines.UnitTests/State/` (3 test files)

The design remains in git history and can be revived if a full scheduler overhaul is ever scheduled.

thomhurst:
## Summary

- remove the unused immutable module state store and related types
- remove its stale DI registration
- remove tests that only covered the dead implementation

## Why

The scheduler and runner still use the mutable ModuleState path. Keeping a second, unwired completion state model creates misleading documentation and duplicate ownership.

## Validation

- `dotnet build test/ModularPipelines.UnitTests/ModularPipelines.UnitTests.csproj --configuration Release --nologo` (passed; existing warnings remain)
- full TUnit executable attempted twice but produced no output and exceeded 60-second and 300-second limits

Closes #2986


thomhurst:
Addressed the public API compatibility concern in `e3b389ad6`:

- Restored `ModuleExecutionPhase` and `ModuleStateSnapshot` with their exact prior public surface.
- Marked both types `[Obsolete]` for removal in the next major version.
- Added a regression test proving both public types remain loadable and obsolete.

Validation:
- Focused compatibility test passes.
- Full unit suite: 878 passed, 6 skipped; the only failure is the existing `GitRootDirectory` assertion that expects the isolated worktree folder to be named exactly `ModularPipelines`.
- Scoped `dotnet format --verify-no-changes` passes.

thomhurst:
Addressed the remaining non-blocking XML documentation concern in 6d8321873. Both retained public compatibility types now explicitly say they are unused legacy models kept for source/binary compatibility and direct readers to the active runtime state types in ModularPipelines.Engine. Release build succeeds with 0 errors; scoped format and diff checks pass.

Issue #3081, opened 2026-07-21
Fix handwritten nullable warnings and unused field in src (CS8600/8602/8604/8620, CS0414)


thomhurst:
From the `switch-mel-spectre` run (https://github.com/thomhurst/ModularPipelines/actions/runs/29871770266/job/88773296619?pr=2619, see #3073) — the small set of warnings in handwritten framework code, easy targeted fixes:

- `src/ModularPipelines/Console/OutputCoordinator.cs(261,28)`: CS8600 — converting null literal or possible null value to non-nullable type.
- `src/ModularPipelines/Engine/ModuleExecutionPipeline.cs(127,58)`: CS8604 — possible null reference argument for parameter 'value' in `ModuleResult<T>.Create...`.
- `src/ModularPipelines/Engine/ModuleExecutionPipeline.cs(236,54)` and `(244,46)`: CS8620 — `ModuleResult<T>` / `ModuleResult<T>.SkippedWrapper` nullability-of-reference-types argument mismatch.
- `src/ModularPipelines.Analyzers/ModularPipelines.Analyzers.CodeFixes/Extensions/SyntaxNodeExtensions.cs(15,51)`: CS8602 — dereference of a possibly null reference.
- `src/ModularPipelines.Distributed.SignalR/Coordination/SignalRWorkerCoordinator.cs(17,27)`: CS0414 — field `_completed` is assigned but never used (either wire it up or delete it — worth checking it isn't a half-finished completion guard).

thomhurst:
Claimed. I am reproducing the five handwritten compiler warnings, applying targeted nullability/state fixes, then rebuilding the affected projects with warning-focused verification.

PR #3095, opened 2026-07-21
fix: generate BuildCommandLine as virtual/override instead of hiding base member (#3080)

### LINKED ISSUE #3080: SourceGenerator CommandOptionsGenerator: BuildCommandLine() hides base member without new (CS0108)
From the `switch-mel-spectre` run (https://github.com/thomhurst/ModularPipelines/actions/runs/29871770266/job/88773296619?pr=2619, see #3073): 94 CS0108 log lines from the Roslyn source generator output (`obj/.../ModularPipelines.SourceGenerator.CommandOptionsGenerator/*.g.cs`):

```
BashCommandOptions.g.cs(14,48): warning CS0108: 'BashCommandOptions.BuildCommandLine()' hides inherited member 'BashOptions.BuildCommandLine()'. Use the new keyword if hiding was intended.
```

Affected generated types (6 warnings each): `BashCommandOptions`, `BashFileOptions`, `PowershellScriptOptions`, `PowershellFileOptions`, and the `AptGet*Options` family — i.e. every case where a derived options class and its base both get a generated `BuildCommandLine()`.

**Fix in `src/ModularPipelines.SourceGenerator` (CommandOptionsGenerator):** when the base type also has a generated `BuildCommandLine()`, emit the member with `new` (or better, generate a `virtual`/`override` pair so the base-typed call sites still dispatch to the derived implementation — worth checking which semantics the command executor actually relies on; if it calls `BuildCommandLine()` through a base reference, the current hiding may be a real bug, not just noise).

thomhurst:
## Summary

Fixes #3080.

`CommandOptionsGenerator` emitted `public ModularPipelines.Models.CommandLine BuildCommandLine()` unconditionally. In `[CliTool]` hierarchies (`BashOptions` → `BashCommandOptions`, `AptGetOptions` → `AptGetInstallOptions`, ...) both levels get a generated method, so the derived one hid the base one:

```
BashCommandOptions.g.cs(14,48): warning CS0108: 'BashCommandOptions.BuildCommandLine()' hides inherited member 'BashOptions.BuildCommandLine()'. Use the new keyword if hiding was intended.
```

Beyond the noise, hiding means a base-typed reference statically binds to the base implementation — wrong dispatch the moment anything consumes these methods polymorphically (nothing in-repo calls them yet; the property-extraction side of this generator is still stubbed).

## Change

`GetBuildMethodModifier` walks the base-type chain and picks:

| Situation | Emitted modifier |
|---|---|
| No ancestor gets a generated method | `virtual` (omitted if the class is `sealed`) |
| Ancestor in the same compilation will get one (predicted via the same partial + `[CliTool]`/property checks `GenerateCode` uses) | `override` |
| Ancestor in a referenced assembly already exposes a parameterless `BuildCommandLine` in metadata | `override` if that method is virtual/override; `new` as a compile-safe fallback for assemblies built with the old generator |

`OptionsClassInfo` carries the modifier as a string, so incremental-generator caching semantics are unchanged (still a value-equatable record).

## Verification

- `dotnet build src/ModularPipelines -c Release --no-incremental`: CS0108 count drops 30 → 0, no errors, no new warnings (remaining CS8600/8602/8604/8620 are the handwritten-code set tracked by #3081).
- Reflection on the built assembly confirms real overriding, not re-hiding:
  - `BashOptions.BuildCommandLine` — declared on `BashOptions`, virtual
  - `BashCommandOptions.BuildCommandLine` — declared on `BashCommandOptions`, `GetBaseDefinition()` = `BashOptions`

### DIFF
--- a/src/ModularPipelines.SourceGenerator/BuildMethodGenerator.cs
+++ b/src/ModularPipelines.SourceGenerator/BuildMethodGenerator.cs
@@ -56,7 +56,7 @@ internal static class BuildMethodGenerator
         sb.AppendLine("    /// <summary>");
         sb.AppendLine("    /// Builds the command line arguments from this options instance.");
         sb.AppendLine("    /// </summary>");
-        sb.AppendLine("    public ModularPipelines.Models.CommandLine BuildCommandLine()");
+        sb.AppendLine($"    public {info.BuildMethodModifier}ModularPipelines.Models.CommandLine BuildCommandLine()");
         sb.AppendLine("    {");
         sb.AppendLine("        var args = new System.Collections.Generic.List<string>();");
 
--- a/src/ModularPipelines.SourceGenerator/CommandOptionsGenerator.cs
+++ b/src/ModularPipelines.SourceGenerator/CommandOptionsGenerator.cs
@@ -130,15 +130,29 @@ public sealed class CommandOptionsGenerator : IIncrementalGenerator
     /// </summary>
     private static string GetBuildMethodModifier(INamedTypeSymbol type, Compilation compilation)
     {
+        var commandLineType = compilation.GetTypeByMetadataName("ModularPipelines.Models.CommandLine");
+
         for (var current = type.BaseType; current is not null; current = current.BaseType)
         {
             // Ancestor compiled in another assembly: its generated method is visible as metadata.
+            // Only methods a derived type in another assembly can actually hide or override matter:
+            // instance, parameterless, non-generic, and at least protected.
             var existing = current.GetMembers("BuildCommandLine")
                 .OfType<IMethodSymbol>()
-                .FirstOrDefault(m => m.Parameters.IsEmpty && !m.IsStatic);
+                .FirstOrDefault(m => m.Parameters.IsEmpty
+                    && !m.IsStatic
+                    && !m.IsGenericMethod
+                    && m.DeclaredAccessibility is Accessibility.Public or Accessibility.Protected or Accessibility.ProtectedOrInternal);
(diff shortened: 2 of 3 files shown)


thomhurst:
Both review findings addressed in the follow-up commit:

1. **Sealed override → CS0239**: `existing.IsSealed` now forces the `new` fallback instead of `override`.
2. **Incompatible overload → broken override**: the metadata lookup now skips generic, static, and less-than-protected members entirely, and only emits `override` when the slot returns `ModularPipelines.Models.CommandLine` and is genuinely overridable (`virtual`/`abstract`/`override`, not sealed). Any other accessible name match gets `new`; inaccessible matches fall through to the normal walk (a private base method can't be hidden, so plain `virtual` is correct there).

Re-verified: core builds clean (no CS0108, no new warnings) and reflection still shows `BashCommandOptions.BuildCommandLine` overriding `BashOptions`' virtual slot.

thomhurst:
Codex finding addressed: the `new` fallback now emits `new virtual` (plain `new` only when the hiding type is sealed), so a generated descendant of that class can emit `override` against a real slot instead of failing with CS0506. Core rebuild stays clean — no CS0108, no new warnings.

thomhurst:
Addressed: `override` is now restricted to public metadata slots (an override can't widen accessibility, so a protected virtual `BuildCommandLine` in a referenced assembly would have hit CS0507). Protected/protected-internal matches take the `new virtual` hiding path instead. Core rebuild clean — no CS0108, no errors, no new warnings.

Issue #3179, opened 2026-07-26
CI: pipeline (ubuntu-latest) master killed during BuildSolutionsModule ("runner received shutdown signal")


thomhurst:
## Symptom

The `pipeline (ubuntu-latest)` job (matrix `instance: 0` = the distributed master) fails partway through the run while `BuildSolutionsModule` is still executing. The other four matrix instances pass in 2–4 min.

Log tail (run [30209099078](https://github.com/thomhurst/ModularPipelines/actions/runs/30209099078/job/89812200128), head `0ac1f583`, failed at ~19 min):

```
Module BuildSolutionOnMacOSModule completed successfully
Scheduler waiting: Total=21, Queued=0, Executing=1, Completed=9, Pending=11
Pending modules: PackProjectsModule, PackagePathsParserModule, RunAllUnitTestsModule,
  RunAnalyzersUnitTestsModule, RunAzureUnitTestsModule, RunCoreUnitTestsModule,
  RunDistributed*UnitTestsModule (x5) ...
Executing modules: BuildSolutionsModule
##[error]The runner has received a shutdown signal. This can happen when the runner
service is stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
```

The pending modules all depend on `BuildSolutionsModule`, which is the single module still running when the runner is terminated.

## What it is NOT

- **Not the distributed result-wait hang** (#3174, fixed in #3178). No `WaitForResultAsync` / coordination stall appears in the log — the master is inside the local `BuildSolutionsModule` build.
- **Not the job timeout**: `timeout-minutes: 60` is set (via #3177); the job died at ~19 min. (Earlier runs died at ~50 min — the duration varies.)
- **Not `concurrency` auto-cancel**: `.github/workflows/dotnet.yml` has no `concurrency` block.

## Scope / history

- Reproduces consistently on this job, including on commit `8cd3a729` (the core-solution-split PR's first commit), which made **no** changes to `BuildSolutionsModule` — so it is not introduced by the recent solution-split / distributed work.
- The `.NET` workflow runs **only on `pull_request`** (no push-to-main trigger), so `main` is validated pre-merge; this check has been red across the recently merged PRs (#3173/#3176/#3177/#3178).

## Likely causes (unconfirmed)

[...]

PR #3182, opened 2026-07-26
Fix/diagnose CI master reclaim: per-solution logging + cap MSBuild parallelism (#3179)

thomhurst:
Targets #3179 (the `pipeline (ubuntu-latest)` master CI blocker).

## Refined diagnosis

The master is killed with `The runner has received a shutdown signal / The operation was canceled` during `BuildSolutionsModule`, at **wildly varying times (~2.5, 14, 19, 50 min)**. A 2.5-min reclaim rules out OOM from the heavy `All.sln` build. That variance fits **runner-agent heartbeat starvation**: `dotnet build` saturates every core, the GitHub runner agent misses its heartbeat, and GitHub reclaims the runner (which is exactly the message it emits). Output is buffered at `Normal` verbosity, so the build logs nothing before the kill — hence the opaque gap.

## Changes (both in `BuildSolutionsModule`)

1. **Per-solution progress logging** — build each solution in an explicit serial loop logging before/after (`Building solution N/6: X` / `Finished … (exit code)`). These flush immediately, so if it still dies we see exactly which solution and how far it got. Behaviour otherwise unchanged (builds were already serial).
2. **Cap MSBuild parallelism** — pass `-maxcpucount:2` to each `dotnet build` so a core stays free for the runner agent to keep heartbeating.

## Why this shape

The two changes are complementary and self-validating: if starvation is the cause, capping parallelism turns the master job green; if it isn't, the logging pinpoints the stuck solution for the next iteration. Both are low-risk — one is logging-only, the other a single build flag. `-maxcpucount:2` assumes a ~4-core runner; tune if that changes.

Not compiled locally (no .NET SDK in this environment); CI is the validation.

### DIFF
--- a/src/ModularPipelines.Build/Modules/BuildSolutionsModule.cs
+++ b/src/ModularPipelines.Build/Modules/BuildSolutionsModule.cs
@@ -1,4 +1,4 @@
-using EnumerableAsyncProcessor.Extensions;
+using Microsoft.Extensions.Logging;
 using ModularPipelines.Attributes;
 using ModularPipelines.Context;
 using ModularPipelines.DotNet.Extensions;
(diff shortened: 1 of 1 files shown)


thomhurst:
Closing this — neither change achieved its goal:

- **`-maxcpucount:2` cap: disproven.** This run was still reclaimed ~13 min into `BuildSolutionsModule`, while no-cap runs survived longer the same round. The reclaim timing is dominated by intermittency, not build parallelism, so the cap only slows the build without preventing the kill.
- **Per-solution logging: doesn't surface.** Module output is buffered and only flushed when the module completes; `BuildSolutionsModule` is killed before completing, so the new `Building solution N/6` lines never appear (raised as a separate observability issue).

Conclusion and recommendations moved to #3179 — it's a runner/infra reclaim, not a pipeline code fix. Closing rather than merging a change with no proven benefit that would also slow every CI build.

---

PR #3190, opened 2026-07-26
Close standalone-execution gaps for OS-only modules

### LINKED ISSUE #3186: Follow-up: close remaining bare-`[RequiresCapability]` standalone-execution gaps from #3117
Non-blocking follow-ups identified during review of #3117 (which fixed the overlapping cross-platform build failure, #3084). #3117 restored `[RunOnWindowsOnly]`/`[RunOnMacOSOnly]` on the platform build modules so **standalone / non-distributed** runs still gate them locally, while OS capability routing continues to work for distributed workers. The same latent pattern remains in a few other places.

### 1. `RunUnitTestModule` uses bare `[RequiresCapability("linux")]` without `[RunOnLinuxOnly]`
`src/ModularPipelines.Build/Modules/UnitTests/RunUnitTestModule.cs:17`

```csharp
[DependsOn<BuildSolutionsModule>(Optional = true)]
[ConsumesArtifact(typeof(BuildSolutionsModule), "build-output", RestorePath = "../../")]
[RequiresCapability("linux")]   // <-- no [RunOnLinuxOnly]
public abstract class RunUnitTestModule : Module<CommandResult>
```

`[RequiresCapability]` only routes work **in distributed mode**. In a plain local run or a non-distributed CI shard, nothing gates this module by OS, so it (and its subclasses, which inherit the attribute) would attempt to run on Windows/macOS too — the same class of bug #3117 closed for the platform build modules. Adding `[RunOnLinuxOnly]` alongside the capability keeps distributed routing working (the OS capability is auto-derived from `RunOn*Only`) while also gating standalone execution.

### 2. Docs teach the unsafe bare-capability pattern
`docs/docs/distributed/capabilities.md:130` (the "Mixed Pipeline" example)

The example gates its Linux module with `[RunOnLinuxOnly]` (line 125, correct) but its Windows module with bare `[RequiresCapability("windows")]` (line 130). A contributor copying the Windows half would reintroduce this bug class in a new module. Make the example consistent — pair `[RunOnWindowsOnly]` with (or in place of) the bare capability, mirroring the Linux half.

### 3. (Low priority, architectural) Two independent `RunOn*Only` → OS-capability maps
`ModuleConditionHandler.GetOperatingSystemConditionTarget` and `DistributedWorkPublisher.CreateAssignment` now each map the `RunOn*Only` attributes to an OS string (`"windows"`/`"linux"`/`"macos"`) independently. They must stay in lockstep if a new OS is ever added. A single shared static helper would make forking structurally impossible. Not worth blocking on, but a cheap hardening.

Items 1 and 2 are the actionable fixes; item 3 is optional cleanup.

_Surfaced by automated review on #3117; verified against current `main` before filing._

thomhurst:
Follow-up to #3117. Three related fixes so OS gating behaves consistently across standalone and distributed execution.

### 1. `RunUnitTestModule`: add `[RunOnLinuxOnly]`
`[RequiresCapability("linux")]` only routes work **in distributed mode**. Without `[RunOnLinuxOnly]`, a plain local run or a non-distributed CI shard would attempt these Linux-only test modules (and their subclasses, which inherit the attribute) on Windows/macOS too — the same standalone-execution gap #3117 closed for the platform build modules. The auto-derived `linux` capability from `[RunOnLinuxOnly]` matches the existing `[RequiresCapability("linux")]`, so distributed routing is unchanged.

### 2. Docs: make the Mixed Pipeline example consistent
`docs/docs/distributed/capabilities.md` gated its Linux module with `[RunOnLinuxOnly]` but its Windows module with a bare `[RequiresCapability("windows")]` — teaching the unsafe pattern a contributor would copy. The Windows half now uses `[RunOnWindowsOnly]` (which auto-derives the `windows` capability), matching the Linux half and gating standalone execution too.

### 3. Deduplicate the `RunOn*Only` → OS-capability mapping
`ModuleConditionHandler` and `DistributedWorkPublisher` each mapped the OS-only attributes to `"windows"`/`"linux"`/`"macos"` independently and had to stay in lockstep as new operating systems were added. Both now use a single shared `OperatingSystemConditions.GetTarget` helper. Behaviour is unchanged — the publisher's three explicit `if` checks become one loop over the module's mandatory conditions using the same mapping.

## Notes
- Behaviour-preserving for the distributed path; the only functional change is that Linux-only unit-test modules are now correctly skipped in **standalone** execution on non-Linux hosts.
- These are the actionable items from #3186. The runtime OS detector (`OsCapabilityDetector`, which maps the *current* host OS) is a separate concern and left as-is.

Closes #3186

---

### DIFF
--- a/docs/docs/distributed/capabilities.md
+++ b/docs/docs/distributed/capabilities.md
@@ -126,8 +126,8 @@ public class RestoreModule : Module<string> { ... }
 [DependsOn<RestoreModule>]
 public class LinuxBuildModule : Module<string> { ... }
 
-// Only on Windows workers
-[RequiresCapability("windows")]
+// Only on Windows workers (auto-detected from [RunOnWindowsOnly])
+[RunOnWindowsOnly]
 [DependsOn<RestoreModule>]
 public class WindowsBuildModule : Module<string> { ... }
 
--- a/src/ModularPipelines.Build/Modules/UnitTests/RunUnitTestModule.cs
+++ b/src/ModularPipelines.Build/Modules/UnitTests/RunUnitTestModule.cs
@@ -15,6 +15,7 @@ namespace ModularPipelines.Build.Modules.UnitTests;
 
 [DependsOn<BuildSolutionsModule>(Optional = true)]
 [ConsumesArtifact(typeof(BuildSolutionsModule), "build-output", RestorePath = "../../")]
+[RunOnLinuxOnly]
 [RequiresCapability("linux")]
 public abstract class RunUnitTestModule : Module<CommandResult>
 {
--- a/src/ModularPipelines/Attributes/OperatingSystemConditions.cs
+++ b/src/ModularPipelines/Attributes/OperatingSystemConditions.cs
@@ -1,0 +1,36 @@
+namespace ModularPipelines.Attributes;
+
+/// <summary>
+/// Single source of truth mapping the OS-only mandatory run conditions
+/// (<see cref="RunOnWindowsOnlyAttribute"/>, <see cref="RunOnLinuxOnlyAttribute"/>,
+/// <see cref="RunOnMacOSOnlyAttribute"/>) to their operating-system capability identifier.
+/// Both the condition handler (which decides whether to defer an OS condition to a worker)
+/// and the distributed work publisher (which stamps the OS capability onto an assignment)
+/// consume this, so the two paths cannot drift as new operating systems are added.
+/// </summary>
+internal static class OperatingSystemConditions
+{
+    /// <summary>Capability identifier for Windows-only modules.</summary>
+    public const string Windows = "windows";
+
+    /// <summary>Capability identifier for Linux-only modules.</summary>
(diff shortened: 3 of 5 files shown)


thomhurst:
CI status: every check is green except `pipeline (ubuntu-latest)` (the instance-0/master shard), which failed with `##[error]The runner has received a shutdown signal` ~2 minutes into `BuildSolutionsModule` — the same #3179 infrastructure reclaim during the memory-heavy full-solution build, not a failure introduced by this PR. The pipeline app compiled and ran all preceding modules (`NugetVersionGeneratorModule`, `FormatMarkdownModule`, etc.) successfully before the runner was reclaimed.

Notably, this change *was* exercised by CI: the worker shards `pipeline (ubuntu-latest, 3)`, `(ubuntu-latest, 4)`, `(windows-latest, 1)` and `(macos-latest, 2)` ran the unit-test modules and passed, `Analyze (csharp)` built the core, and both automated reviews are clean (Greptile 5/5, no findings).

---

PR #3192, opened 2026-07-26
Revalidate the runnable dependency graph at run time

### LINKED ISSUE #3189: Revalidate the canonical dependency graph at run time (dynamic/registration-event deps + condition-dependent runnability)
Follow-up to the two unresolved **P2** review comments on #3086 (both on `src/ModularPipelines/PipelineImpl.cs`). Captured here because a correct fix is an architectural change, not the one-liner the review threads imply — see the hazard section.

## The two review comments

1. **Revalidate modules discovered for execution** ([#3086 review](https://github.com/thomhurst/ModularPipelines/pull/3086#discussion_r3653026884)): When full-graph validation fails because a conditionally-ignored module has a missing fluent dependency, `PipelineImpl.CreateAsync` falls back to validating only the build-time discovery snapshot (`ModuleRetriever.GetRunnableModulesForValidation()`, which is uncached). `PipelineInitializer` evaluates discovery **again** during `RunAsync` after pipeline-start hooks, so a stateful / time-varying run condition can make a module runnable at run time whose (missing) deps were never revalidated — it then fails late in `DependencyWaiter`.

2. **Populate dynamic dependencies before validating** ([#3086 review](https://github.com/thomhurst/ModularPipelines/pull/3086#discussion_r3653119112)): `IModuleRegistrationEventReceiver` implementations add edges via `IModuleRegistrationContext.AddDependency`, but receivers are only invoked during `ModuleExecutor.InitializeSchedulerAsync` in `RunAsync`. So `BuildAsync`/`ValidateAsync` (and `CreateAsync`) see an empty `IModuleDependencyRegistry` and never revalidate afterward. A registration-time missing dependency or cycle passes validation and only fails at run time.

## Root cause

Validation runs at **build time**, but the true executable graph — the runnable set after pipeline-start hooks + discovery, plus dynamic edges added by registration receivers — is only known at **run time**. The successful path never revalidates. Both reviewers explicitly accept *"revalidate the canonical graph immediately after those events run"* as a remedy.

## Lifecycle (verified against current `main`)

Run-time order inside `RunAsync`:

1. `ExecutionOrchestrator.ExecuteInternal` → `PipelineInitializer.Initialize` — `OnPipelineStartAsync` (hooks) → `ModuleRetriever.GetOrganizedModules` (discovery / `ShouldIgnore`, **cached**) → dependency-chain init. `OrganizedModules` exposes `RunnableModules`, `IgnoredModules`, `AllModules`.
2. `PipelineExecutor.ExecuteAsync(runnableModules, organizedModules)` → `ModuleExecutor.ExecuteAsync(runnableModules)`.
3. `ModuleExecutor.InitializeSchedulerAsync` (`src/ModularPipelines/Engine/ModuleExecutor.cs:95`): `InvokeRegistrationEventsAsync(modules)` (**dynamic edges written to the `IModuleDependencyRegistry` singleton here**) → `scheduler.InitializeModules(modules)`.
4. Missing non-optional edges currently surface late in `DependencyWaiter.WaitForDependenciesAsync` (`src/ModularPipelines/Engine/Execution/DependencyWaiter.cs:45`).

`DistributedModuleExecutor.ExecuteAsync` (`src/ModularPipelines/Distributed/Master/DistributedModuleExecutor.cs:62,71`) mirrors this order. `ModuleDependencyValidator.Validate(modules, dynamicRegistry, metadataRegistry)` (`src/ModularPipelines/Modules/ModuleDependencyValidator.cs:22`) consults the dynamic registry via `ModuleDependencyResolver.GetAllDependencies` (`ModuleDependencyResolver.cs:214-221`), so it *can* see dynamic edges once they are populated.

## ⚠️ Hazard: the obvious fix is wrong

The tempting fix — call `ModuleDependencyValidator.Validate(modules, _dependencyRegistry, _metadataRegistry)` in `ModuleExecutor.InitializeSchedulerAsync` right after `InvokeRegistrationEventsAsync` — introduces a **false-positive regression**:

`ModuleDependencyValidator.Validate(modules, …)` treats its `modules` argument as **both** "the modules to validate" **and** "the universe of registered module types" (`moduleTypes = registeredModules...ToHashSet()`; a required dep is "missing" if `!moduleTypes.Contains(dependencyType)`).

- Build-time validation passes the **full** registered `IModule` set, so an OS-skipped dependency still counts as registered.
- Passing only the **runnable** set means a runnable module with a required `[DependsOn]` on a **registered-but-conditionally-skipped** module (e.g. a cross-platform module depending on a `[RunOnLinuxOnly]` module, run on Windows/macOS or non-distributed) would throw a spurious `ModuleNotRegisteredException` — behavior that currently does *not* error.

So validating the runnable set *against itself* is incorrect.

## Correct design

Validate the **runnable set's dependencies** against the **full discovered set** (`RunnableModules` ∪ `IgnoredModules`, i.e. everything registered), with the dynamic registry populated:

- Modules to validate: the runnable set (matches execution; avoids re-flagging deliberately-ignored modules with intentionally-absent platform deps — the case `CreateAsync`'s fallback already handles).
- "Available" universe: all discovered/registered module types (so a skipped-but-registered dependency is *not* missing).
- Only genuinely-unregistered targets (including dynamic ones) throw; cycles among the runnable graph are detected.

This requires either:

- **(a)** a new `ModuleDependencyValidator` overload separating *modules-to-validate* from *available-types*, plus threading the full discovered set (`OrganizedModules.AllModules`) into `ModuleExecutor` / `DistributedModuleExecutor` (their `IModuleExecutor.ExecuteAsync` only receives the runnable set today); or
- **(b)** relocating registration-event invocation earlier (e.g. into `PipelineInitializer`, where the full set is available) and validating there — removing the invocation from both executors to avoid double-invoking receivers (whose `OnRegistrationAsync` is arbitrary, possibly non-idempotent, user code).

Watch-outs: invoke registration receivers **exactly once**; don't re-run pipeline-start hooks or a third `ShouldIgnore` pass; keep `WorkerModuleExecutor` (worker side) out of scope — it executes individually assigned modules and does no graph validation.

## Suggested regression tests

- Registration receiver adds a dependency on an **unregistered** module (not declared via `[DependsOn]`, so not auto-registered) → expect an eager validation error rather than a late `DependencyWaiter` failure. (Extends `test/ModularPipelines.UnitTests/Attributes/DynamicDependencyIntegrationTests.cs`.)
- Registration receiver introduces a **cycle** → eager `DependencyCollisionException`.
- **Guard against the hazard:** a runnable module with a required `[DependsOn]` on a `[RunOnLinuxOnly]`/`[RunOnWindowsOnly]` module that is skipped on the current OS → must **not** throw (regression test for the false-positive above).
- A stateful run condition that flips a module to runnable after build-time validation, whose dependency is missing → caught at run time.

## Notes

- The build-time paths (`PipelineImpl.CreateAsync`, `Validation/DependencyValidator.cs`) can't see dynamic/condition-dependent edges regardless, so this is fundamentally run-time validation; `BuildAsync`/`ValidateAsync` will still not catch these unless registration receivers are invoked at build time (option (b) territory).
- Attempted during a follow-up to #3086; reverted before landing because a correct fix needs the architectural change above and could not be validated locally (full build/test suite unavailable in that environment).

thomhurst:
Addresses the two open **P2** review comments from #3086 (both on `PipelineImpl.cs`).

## Problem

Dependency validation runs at build time (`BuildAsync`/`ValidateAsync` → `PipelineImpl.CreateAsync`), but two kinds of edges are only known at **run time**:

1. **Dynamic dependencies** added via `IModuleRegistrationContext.AddDependency` — the registration-event receivers that add them are only invoked during `RunAsync` (`ModuleExecutor.InitializeSchedulerAsync`), so build-time validation sees an empty dependency registry.
2. **Condition-dependent runnability** — `PipelineInitializer` re-evaluates discovery during `RunAsync` after pipeline-start hooks, so a stateful/time-varying run condition can make a module runnable whose dependencies were never validated.

The successful path never revalidated afterward, so a registration-time missing/cyclic dependency passed validation and only failed late in the scheduler's `DependencyWaiter`.

## Fix

After registration events populate dynamic dependencies (and discovery has settled), the standalone `ModuleExecutor` and the master `DistributedModuleExecutor` revalidate the **runnable set** — `ModuleDependencyValidator.Validate(modules, dynamicRegistry, metadataRegistry)` — before scheduling.

This is an **eager mirror of the existing `DependencyWaiter` check**: the validator throws for a required dependency that is not in the runnable set, which is the identical condition under which `DependencyWaiter` throws `ModuleNotRegisteredException` (a required dependency whose `GetModuleCompletionTask` is null because it isn't a scheduled/runnable module). So validation and runtime agree — the check simply runs earlier, before scheduling, with a clearer message — and dynamic missing/cyclic dependencies are caught eagerly instead of surfacing late.

## Changes

- `ModuleExecutor` / `DistributedModuleExecutor`: inject the dependency + metadata registries; validate the runnable set immediately after registration events, before scheduling. (`WorkerModuleExecutor` is intentionally untouched — workers execute individually-assigned modules and don't build the graph.)
- `ModuleDependencyValidator`: **unchanged** — the fix reuses the existing overload. (An earlier revision added a separate available-universe overload; that was dropped after review, since it made the validator accept a dependency the runtime rejects.)
- Tests:
  - End-to-end: a registration-event dependency on an unregistered module now fails fast with `ModuleNotRegisteredException` during `RunAsync` instead of surfacing late.
  - Unit: validation over the dynamic dependency registry throws for a dynamic dependency outside the validated set and passes when it's present.

## Notes
- This is run-time revalidation (the reviewers' explicitly-accepted remedy: *"revalidate the canonical graph immediately after those events run"*). `BuildAsync`/`ValidateAsync` still can't see dynamic/condition-dependent edges, since those genuinely don't exist until run time.
- Behaviour change: a dynamic missing/cyclic dependency that previously failed late now fails eagerly during `RunAsync` — the intended fix. Whether a module depending on a *skipped* module should cascade-skip rather than error is a separate, pre-existing `DependencyWaiter` behaviour and out of scope here.

Closes #3189

### DIFF
--- a/src/ModularPipelines/Distributed/Master/DistributedModuleExecutor.cs
+++ b/src/ModularPipelines/Distributed/Master/DistributedModuleExecutor.cs
@@ -7,6 +7,7 @@ using ModularPipelines.Distributed.Serialization;
 using ModularPipelines.Distributed.Worker;
 using ModularPipelines.Engine;
 using ModularPipelines.Engine.Attributes;
+using ModularPipelines.Engine.Dependencies;
 using ModularPipelines.Engine.Execution;
 using ModularPipelines.Models;
 using ModularPipelines.Modules;
@@ -24,6 +25,9 @@ internal class DistributedModuleExecutor(
     ModuleTypeRegistry typeRegistry,
     ModuleResultSerializer serializer,
     IModuleResultRegistry resultRegistry,
+    IEnumerable<IModule> registeredModules,
+    IModuleDependencyRegistry dependencyRegistry,
+    IModuleMetadataRegistry metadataRegistry,
     IOptions<DistributedOptions> options,
     ArtifactLifecycleManager? artifactLifecycleManager,
     ILogger<DistributedModuleExecutor> logger) : IModuleExecutor
@@ -38,6 +42,9 @@ internal class DistributedModuleExecutor(
     private readonly ModuleTypeRegistry _typeRegistry = typeRegistry;
     private readonly ModuleResultSerializer _serializer = serializer;
     private readonly IModuleResultRegistry _resultRegistry = resultRegistry;
+    private readonly IEnumerable<IModule> _registeredModules = registeredModules;
+    private readonly IModuleDependencyRegistry _dependencyRegistry = dependencyRegistry;
+    private readonly IModuleMetadataRegistry _metadataRegistry = metadataRegistry;
     private readonly IOptions<DistributedOptions> _options = options;
     private readonly ArtifactLifecycleManager? _artifactLifecycleManager = artifactLifecycleManager;
     private readonly ILogger<DistributedModuleExecutor> _logger = logger;
@@ -60,6 +67,12 @@ internal class DistributedModuleExecutor(
 
         // Invoke registration events before dependency resolution
         await _registrationEventExecutor.InvokeRegistrationEventsAsync(modules).ConfigureAwait(false);
+
+        // Revalidate the runnable set now that registration-event dependencies are populated, so
+        // missing/self/cyclic dependencies (including ones added via AddDependency) fail fast here
+        // rather than the master hanging or failing late. The full registered set is the available
+        // universe, so depending on a registered-but-skipped module is not misreported as missing.
+        ModuleDependencyValidator.Validate(modules, _registeredModules, _dependencyRegistry, _metadataRegistry);
 
         // Wait for workers to register before distributing work
         await WaitForWorkersAsync(_lifetime.ApplicationStopping);
--- a/src/ModularPipelines/Distributed/Master/DistributedModuleExecutor.cs
+++ b/src/ModularPipelines/Distributed/Master/DistributedModuleExecutor.cs
@@ -25,7 +25,6 @@ internal class DistributedModuleExecutor(
     ModuleTypeRegistry typeRegistry,
     ModuleResultSerializer serializer,
     IModuleResultRegistry resultRegistry,
-    IEnumerable<IModule> registeredModules,
     IModuleDependencyRegistry dependencyRegistry,
     IModuleMetadataRegistry metadataRegistry,
     IOptions<DistributedOptions> options,
@@ -42,7 +41,6 @@ internal class DistributedModuleExecutor(
     private readonly ModuleTypeRegistry _typeRegistry = typeRegistry;
     private readonly ModuleResultSerializer _serializer = serializer;
     private readonly IModuleResultRegistry _resultRegistry = resultRegistry;
-    private readonly IEnumerable<IModule> _registeredModules = registeredModules;
     private readonly IModuleDependencyRegistry _dependencyRegistry = dependencyRegistry;
     private readonly IModuleMetadataRegistry _metadataRegistry = metadataRegistry;
     private readonly IOptions<DistributedOptions> _options = options;
@@ -70,9 +68,8 @@ internal class DistributedModuleExecutor(
 
         // Revalidate the runnable set now that registration-event dependencies are populated, so
         // missing/self/cyclic dependencies (including ones added via AddDependency) fail fast here
-        // rather than the master hanging or failing late. The full registered set is the available
-        // universe, so depending on a registered-but-skipped module is not misreported as missing.
-        ModuleDependencyValidator.Validate(modules, _registeredModules, _dependencyRegistry, _metadataRegistry);
+        // rather than the master hanging or failing late. Mirrors the standalone ModuleExecutor.
+        ModuleDependencyValidator.Validate(modules, _dependencyRegistry, _metadataRegistry);
 
         // Wait for workers to register before distributing work
         await WaitForWorkersAsync(_lifetime.ApplicationStopping);
--- a/src/ModularPipelines/Engine/ModuleExecutor.cs
+++ b/src/ModularPipelines/Engine/ModuleExecutor.cs
@@ -30,7 +30,6 @@ internal class ModuleExecutor : IModuleExecutor
     private readonly IParallelLimitProvider _parallelLimitProvider;
     private readonly IRegistrationEventExecutor _registrationEventExecutor;
     private readonly IMetricsCollector _metricsCollector;
-    private readonly IEnumerable<IModule> _registeredModules;
     private readonly IModuleDependencyRegistry _dependencyRegistry;
     private readonly IModuleMetadataRegistry _metadataRegistry;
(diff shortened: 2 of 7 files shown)


thomhurst:
Thanks @greptile-apps — that finding was correct and I've reworked the approach in `4a286d3`.

I traced the engine to confirm the concern: `DependencyWaiter.WaitForDependenciesAsync` iterates `moduleState.Dependencies` and throws `ModuleNotRegisteredException` when `scheduler.GetModuleCompletionTask(dep)` returns null (i.e. the dependency isn't in the runnable set) for a non-optional dependency. So a runnable module with a required dependency on a skipped module **already fails at run time today** — there's no graceful handling, so the "registered-but-skipped" case the original overload tried to accommodate isn't actually a false positive; it's a genuine failure the runtime rejects. The available-universe overload made the validator *accept* something the runtime *rejects* — exactly the mismatch you flagged.

**Change:** dropped the extra overload and the registered-module-set injection. The executors now call the existing `ModuleDependencyValidator.Validate(modules, dynamicRegistry, metadataRegistry)` on the runnable set. Its throw condition (`required dependency not in the validated set`) is identical to `DependencyWaiter`'s (`required dependency has no completion task`, i.e. not in the runnable set), so validation and runtime now agree — the check just runs eagerly, before scheduling, with a clearer message. `ModuleDependencyValidator` itself is unchanged; the net diff is ~26 lines across the two executors.

Whether a module depending on a *skipped* module should cascade-skip rather than error is a separate, pre-existing behaviour (in `DependencyWaiter`) and out of scope here.

---

thomhurst:
CI status on `4a286d3`: every check is green except `pipeline (ubuntu-latest)` (the instance-0/master shard), which failed with `##[error]The runner has received a shutdown signal` ~7 minutes into `BuildSolutionsModule` — the recurring #3179 infrastructure reclaim during the memory-heavy full-solution build, not a failure from this PR. (`claude-review` shows `cancelled`, also not a failure.)

Evidence this change is sound:
- The master pipeline ran **7 modules to completion** and had others executing before the runner was reclaimed — so the new `ModuleDependencyValidator.Validate(...)` call in `InitializeSchedulerAsync` passed on the real build pipeline's runnable graph (no false rejection).
- The worker shards `pipeline (ubuntu-latest, 3)`, `(ubuntu-latest, 4)`, `(windows-latest, 1)` and `(macos-latest, 2)` ran the updated unit + integration tests and passed; `Analyze (csharp)` built the core; Greptile is 5/5.

---

PR #3193, opened 2026-07-26
Revalidate canonical dependency graph at runtime

### LINKED ISSUE #3189: Revalidate the canonical dependency graph at run time (dynamic/registration-event deps + condition-dependent runnability)

thomhurst:
## Summary
- revalidate runnable modules after registration receivers populate dynamic dependencies
- validate against the full discovered module universe while detecting cycles only in the runnable graph
- apply the same eager validation on distributed masters; workers remain unchanged
- cover missing dynamic edges, dynamic cycles, changing run conditions, target/universe separation, and master shutdown signaling

## Validation
- core project build: 0 errors
- DynamicDependencyIntegrationTests: 5/5 passed via minimal focused harness
- DistributedModuleExecutorTests: 17/17 passed

Closes #3189

### DIFF
--- a/src/ModularPipelines/Distributed/Master/DistributedModuleExecutor.cs
+++ b/src/ModularPipelines/Distributed/Master/DistributedModuleExecutor.cs
@@ -7,6 +7,7 @@ using ModularPipelines.Distributed.Serialization;
 using ModularPipelines.Distributed.Worker;
 using ModularPipelines.Engine;
 using ModularPipelines.Engine.Attributes;
+using ModularPipelines.Engine.Dependencies;
 using ModularPipelines.Engine.Execution;
 using ModularPipelines.Models;
 using ModularPipelines.Modules;
@@ -18,6 +19,8 @@ internal class DistributedModuleExecutor(
     IModuleSchedulerFactory schedulerFactory,
     IModuleRunner moduleRunner,
     IRegistrationEventExecutor registrationEventExecutor,
+    IModuleDependencyRegistry dependencyRegistry,
+    IModuleMetadataRegistry metadataRegistry,
     IDistributedCoordinator coordinator,
     DistributedWorkPublisher publisher,
     DistributedResultCollector resultCollector,
@@ -32,6 +35,8 @@ internal class DistributedModuleExecutor(
     private readonly IModuleSchedulerFactory _schedulerFactory = schedulerFactory;
     private readonly IModuleRunner _moduleRunner = moduleRunner;
     private readonly IRegistrationEventExecutor _registrationEventExecutor = registrationEventExecutor;
+    private readonly IModuleDependencyRegistry _dependencyRegistry = dependencyRegistry;
+    private readonly IModuleMetadataRegistry _metadataRegistry = metadataRegistry;
     private readonly IDistributedCoordinator _coordinator = coordinator;
     private readonly DistributedWorkPublisher _publisher = publisher;
     private readonly DistributedResultCollector _resultCollector = resultCollector;
@@ -42,7 +47,9 @@ internal class DistributedModuleExecutor(
     private readonly ArtifactLifecycleManager? _artifactLifecycleManager = artifactLifecycleManager;
     private readonly ILogger<DistributedModuleExecutor> _logger = logger;
 
-    public async Task<IEnumerable<IModule>> ExecuteAsync(IReadOnlyList<IModule> modules)
+    public async Task<IEnumerable<IModule>> ExecuteAsync(
+        IReadOnlyList<IModule> modules,
+        IReadOnlyList<IModule> availableModules)
     {
         if (modules.Count == 0)
         {
@@ -58,15 +65,21 @@ internal class DistributedModuleExecutor(
         // Build O(1) lookup for module resolution
         var moduleLookup = DependencyResultApplicator.BuildModuleLookup(modules);
 
-        // Invoke registration events before dependency resolution
-        await _registrationEventExecutor.InvokeRegistrationEventsAsync(modules).ConfigureAwait(false);
-
-        // Wait for workers to register before distributing work
-        await WaitForWorkersAsync(_lifetime.ApplicationStopping);
-
         IModuleScheduler? scheduler = null;
         try
         {
+            // Invoke registration events before dependency resolution
+            await _registrationEventExecutor.InvokeRegistrationEventsAsync(modules).ConfigureAwait(false);
+
+            ModuleDependencyValidator.Validate(
+                modules,
+                availableModules,
+                _dependencyRegistry,
+                _metadataRegistry);
+
+            // Wait for workers to register before distributing work
+            await WaitForWorkersAsync(_lifetime.ApplicationStopping);
+
             scheduler = _schedulerFactory.Create();
             scheduler.InitializeModules(modules);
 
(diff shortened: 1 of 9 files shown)


thomhurst:
@codex review

Exact validated head: 14493df06c87e6717210ac8f7bfa2089fb3c7845. Local gates: core build (0 errors), DynamicDependencyIntegrationTests 5/5, DistributedModuleExecutorTests 17/17, changed-file whitespace verification.

thomhurst:
Closing as a duplicate of #3192, which merged while this parallel implementation was being prepared. No further action needed here.

PR #3198, opened 2026-07-27
Give the CI pipeline build memory headroom to stop the runner OOM-kill

thomhurst:
## Problem (#3179)

The `pipeline (ubuntu-latest)` (instance 0 / master) check fails on essentially every PR — it's the sole red check that's forced an admin override on the last several merges (#3117, #3190, #3192). The signature is always the same:

```
Executing modules: BuildSolutionsModule
##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
```

It happens **partway through `BuildSolutionsModule`** (which builds the full 60+ project `ModularPipelines.All.sln`), at **variable times** (~2 → ~7 → ~19 → ~50 min across runs), while every other check — worker shards, `Analyze`, reviews — is green. That signature + a memory-heavy full-solution build on a 7 GB `ubuntu-latest` runner is the classic profile of the host **OOM-killing the runner agent**.

## Fix

Two low-risk, workflow-only levers that address the out-of-memory reclaim at its root:

1. **Increase swap** — replace the default 4 GB swapfile with a 10 GB one on `/mnt` (the ephemeral disk, so the root filesystem stays free for restore/build output). 7 GB RAM + 10 GB swap gives the full-solution build room to complete rather than being killed. The step logs `free -h` so the active swap is visible.
2. **`DOTNET_gcServer=0`** — run the pipeline and the `dotnet build`/`test` child processes it spawns under workstation GC. Server GC allocates a heap per core, inflating peak memory on the multi-core runners during the full-solution build; workstation GC keeps the footprint far lower.

No product code changes — only `.github/workflows/dotnet.yml`.

## Why not just cap parallelism

#3182 tried `-maxcpucount:2` on its own and still overshot RAM, because halving concurrency without adding headroom still peaked above 7 GB. Swap provides that headroom, and workstation GC lowers the peak. If this proves insufficient, reducing build parallelism (or a larger runner for `instance: 0`) remains available as a follow-up lever.

## Validation note

This targets the runner environment, so it can only really be proven by CI itself — if instance 0 gets through `BuildSolutionsModule` to completion, the fix holds. The `free -h` output in the new step will confirm the swap is active.

Toward #3179

---

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -66,6 +66,26 @@ jobs:
         uses: samypr100/setup-dev-drive@v4
         with:
           drive-size: 5GB
+      # The distributed master (and other Linux instances) builds the full
+      # ModularPipelines.All.sln, whose peak memory can exceed the 7 GB RAM of a
+      # standard ubuntu-latest runner. When it does, the host OOM-kills the runner
+      # agent, which surfaces as "The runner has received a shutdown signal" mid-build
+      # (see #3179). Add a large swapfile so the build has headroom to complete instead
+      # of being reclaimed.
+      - name: Increase swap space
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          # Replace the default 4 GB swapfile with a larger one on /mnt (the ephemeral
+          # disk), leaving the root filesystem free for the NuGet restore and build output.
+          sudo swapoff -a || true
+          sudo rm -f /mnt/swapfile || true
+          sudo fallocate -l 10G /mnt/swapfile
+          sudo chmod 600 /mnt/swapfile
+          sudo mkswap /mnt/swapfile
+          sudo swapon /mnt/swapfile
+          echo "Memory and swap after reconfiguration:"
+          free -h
       - name: Setup .NET
         uses: actions/setup-dotnet@v6.0.0
         with:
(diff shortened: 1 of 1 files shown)


thomhurst:
Thanks both. A few clarifications, then I'll let CI be the judge:

**On the `DOTNET_gcServer` propagation concern (@greptile-apps P2):** it does reach the child processes. `CliCommandFactory` applies `CommandExecutionOptions.EnvironmentVariables` via CliWrap's `WithEnvironmentVariables`, which *overlays* the inherited parent environment rather than replacing it — the `null` entries in `RunUnitTestModule` simply unset specific vars (the distributed-mode connection vars), so everything else, including `DOTNET_gcServer=0`, is still inherited. And `BuildSolutionsModule` — which is the actual OOM path — sets no custom environment at all, so it inherits the setting outright. So there's no GC gap on the memory-heavy build.

**On scoping the swap step to `matrix.instance == 0` (@claude):** I kept it on all Linux instances deliberately. The heavy `All.sln` build is a distributed module — while instance 0 has been the one to pick it up in practice, there's no hard guarantee a Linux *worker* (instance 3/4) won't, and extra swap on a shard is a ~cheap no-op whereas *missing* swap on whichever instance runs the build reintroduces the failure. Reliability over tidiness here; happy to tighten it to instance 0 later if we confirm the assignment is deterministic.

**On swap thrashing vs. the 60-min timeout (@claude):** agreed, that's the thing to watch — the fix may trade an OOM-kill for a slower build. The new step logs `free -h`, and I'll be watching instance 0's duration. If it creeps toward the ceiling, the follow-up lever is capping MSBuild parallelism (lower peak, less spill) or a larger `instance: 0` runner.

---

thomhurst:
Closing — this was based on a wrong premise. The `free -h` output in the CI run showed the runner has 16 GB RAM and the swap step gave it 9 GB more (≈24 GB total), and it *still* got reclaimed mid-build, while `Analyze (csharp)` compiled the same code on the same runner and passed. So it was never a memory shortage.

The real cause (from the instance-0 log): `Redis discovery preflight failed (HttpRequestException); distributed mode disabled` → the primary instance runs the entire pipeline standalone while the others exit. Redis (Upstash) is down, so distribution never engages. Superseding this with a PR that switches distributed mode off explicitly and addresses the standalone build itself.

---

PR #3201, opened 2026-07-27
Switch CI off distributed mode and bound build memory

### LINKED ISSUE #3179: CI: pipeline (ubuntu-latest) master killed during BuildSolutionsModule ("runner received shutdown signal")

thomhurst:
Replaces #3198. Fixes the persistent `pipeline (ubuntu-latest)` failure (#3179).

## What was actually happening

The instance-0 log revealed the real cause — not OOM:

```
Redis discovery preflight failed (HttpRequestException); distributed mode disabled.
Redis discovery is unavailable; primary instance running standalone.
```

Upstash Redis is unreachable, so `ConfigureDistributedModeAsync` falls back to `ShouldRunStandalone`: the **primary instance runs the entire build+test pipeline alone** (16-way parallelism) while instances 1–4 exit immediately. That lone instance keeps getting reclaimed mid-`BuildSolutionsModule` ("The runner has received a shutdown signal", at variable times). The `free -h` from the runner showed **16 GB RAM** (a prior swap experiment in #3198 gave it 24 GB and it *still* died), and `Analyze (csharp)` compiles the same code on the same runner and passes — so it was never a simple RAM shortage. It was the whole distributed load collapsing onto one runner, plus memory accumulating across the sequential solution builds.

## Change

Per direction (Redis is down), switch CI off distributed mode and bound the build's memory:

- **Standalone per platform.** ubuntu/windows/macos each run as an independent job with `TOTAL_INSTANCES=1`, so the pipeline skips Redis/SignalR discovery entirely and no instance inherits everyone else's work. Removes the two redundant ubuntu worker instances and the cloudflared install (distributed-only). This also **restores windows/macos runs**, which had been exiting without doing work while distribution was broken.
- **`MSBUILDDISABLENODEREUSE=1`** — tear down MSBuild/compiler nodes between `BuildSolutionsModule`'s sequential solution builds so their memory doesn't accumulate across the run (a strong candidate for the variable-timing reclaim).
- **`DOTNET_gcServer=0`** — workstation GC keeps the footprint lower than server GC's per-core heaps.
- **10 GB Linux swap** — extra headroom.

## Honest caveats

- I can't run the build pipeline locally, so this is validated only by CI. The `free -h` output and instance-0 reaching completion will tell us if it holds.
- Switching to standalone alone wouldn't fix it (the old fallback already ran standalone and died) — the memory levers above are what target the reclaim. If instance 0 still dies, the next lever is capping MSBuild parallelism (`-maxcpucount`) or a larger runner.
- **Branch protection:** the `pipeline (ubuntu-latest, 3)` and `(4)` checks no longer exist, so required-status-check settings likely need updating.
- The proper long-term fix for distribution is restoring Upstash Redis reachability; this makes CI green and reliable without it.

Closes #3179

---


thomhurst:
Addressed Claude review cleanup in c7bc97698: removed unused R2 secret masks and documented beside TOTAL_INSTANCES that restoring distributed mode also requires Redis/R2 credentials plus cloudflared because EnableTunnel=true.

thomhurst:
Addressed review follow-ups: stale R2 masks were removed in c7bc97698, and 28372b5ff documents the cloudflared/TOTAL_INSTANCES coupling beside EnableTunnel. Keeping swap as temporary headroom while this run validates the standalone build.

thomhurst:
Re: the concern that `640a7ac0` (restoring default MSBuild parallelism) reverts a necessary fix and risks reintroducing #3179 — this is based on a timing assumption that the CI data contradicts.

The premise is that the reclaim in `e26d69d7` happened *after* `f3c65dd0` removed the snapshot tests. It didn't:

- `f3c65dd0` (snapshot tests removed) authored 08:16:27 UTC.
- `e26d69d7` (serialization) authored **08:17:19 UTC** — before `f3c65dd0`'s "Build solutions" step had even started (08:17:31).

So `e26d69d7`'s "exhausted the runner again even after swap" was reacting to the earlier `2dc2f685` failure, where the all-integrations snapshot project **still existed** (isolated in its own 43-reference project referencing AWS+Azure+Google together). `f3c65dd0` was the first commit where those references were actually gone.

And `f3c65dd0`'s run ([30249269627](https://github.com/thomhurst/ModularPipelines/actions/runs/30249269627)) — default MSBuild parallelism, no serialization — completed **fully green**:

| Step | Result | Time |
|---|---|---|
| Build solutions | success | 08:17:31 → 08:26:55 |
| Run Pipeline (tests) | success | 08:26:55 → 08:35:05 |

So default parallelism does fit on the runner once no single project pulls the three big cloud SDKs into one compilation unit. The serialization was addressing a failure that a later commit had already eliminated by a different means, so `640a7ac0` drops it rather than keeping a redundant mitigation. `640a7ac0`'s own run is in progress and expected to confirm this.

The stale PR-description point (Windows/macOS "restored" vs. actually removed) is fair and worth fixing separately.

---

PR #3204, opened 2026-07-27
Preserve executable parent commands in sub-domains

### LINKED ISSUE #3169: Preserve executable parent commands in shared sub-domain generation
Common service grouping drops parent invocation when a command is both executable and has child commands.

Regression from #3147:

- `KubernetesClusterInfoOptions` models executable `kubectl cluster-info`.
- `IKubernetes` exposes `KubernetesClusterinfo` because `cluster-info dump` exists.
- Sub-domain service exposes only `Dump(...)`; parent command becomes unreachable.
- Shared Markdown generator drops same parent from command table.

This collision pattern exists in many hierarchical CLIs and must be fixed in common generators.

Generic fix:

- Represent command nodes as independently executable even when they own children.
- Update shared service interface/implementation/sub-domain generators to emit collision-safe parent invocation (`Execute`, domain-specific root alias, or equivalent).
- Update `GetNonCollidingRootCommands`/documentation exposure logic so executable parents remain listed.
- Ensure dependency injection/service construction exposes both parent and children.
- Add tool-neutral fixture with `tool group` and `tool group child`, asserting both typed calls and rendered commands.
- Keep kubectl cluster-info as regression fixture.

No Kubernetes-only manual service method.

thomhurst:
## Summary

- match executable parent commands to sub-domains through their normalized generated identifiers
- expose parent commands through `SubDomain.Execute(...)` while keeping child methods and DI construction intact
- include executable parents in generated command documentation
- regenerate Kubernetes so `Clusterinfo.Execute(...)` invokes `kubectl cluster-info`

## Testing

- `dotnet build tools/ModularPipelines.OptionsGenerator/ModularPipelines.OptionsGenerator.sln -c Release -m:1 --no-restore`
- OptionsGenerator tests: 500/500 passed
- `dotnet build src/ModularPipelines.Kubernetes/ModularPipelines.Kubernetes.sln -c Release -m:1 --no-restore`
- scoped `dotnet format whitespace --verify-no-changes`

Closes #3169
## Documentation regeneration note

The kubectl module example changed from `Autoscale` to `Annotate` because regeneration reconciled pre-existing checked-in documentation drift; parent-command selection does not cause that example change.

### DIFF
--- a/docs/docs/mp-packages/cli/kubectl.md
+++ b/docs/docs/mp-packages/cli/kubectl.md
@@ -30,8 +30,8 @@ public class RunCommandModule : Module<CommandResult>
         IModuleContext context,
         CancellationToken cancellationToken)
     {
-        return await context.Kubernetes().Autoscale(
-            new KubernetesAutoscaleOptions(),
+        return await context.Kubernetes().Annotate(
+            new KubernetesAnnotateOptions(),
             cancellationToken: cancellationToken);
     }
 }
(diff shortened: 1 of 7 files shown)


thomhurst:
Addressed both review nice-to-haves in 284e2b137. Duplicate normalized-identifier grouping now uses one shared helper while preserving the two tailored diagnostics. PR description now calls out the unrelated kubectl `Autoscale` → `Annotate` regeneration drift. Full OptionsGenerator suite: 508/508; Release build succeeded.

thomhurst:
Addressed the final diagnostic observation in 98a8e4f47. Duplicate executable-parent failures now report the normalized identifier and every conflicting full CLI command; regression asserts the exact detail. GeneratorHardeningTests: 35/35 passed; scoped format clean.

thomhurst:
Rebased conflict-free onto current `main` (`5eb34cb341`) after mergeability entered `UNKNOWN`. New head: `2b6a204a0`.

`git range-diff` reports all three PR commits patch-equivalent (`=`); worktree is clean. Existing green checks/review will rerun. Per queue policy, not merging in this push iteration.

PR #3206, opened 2026-07-27
Emit only safe runnable CLI examples

### LINKED ISSUE #3167: Make shared documentation generator emit safe runnable examples
`MarkdownDocumentationGenerator.AppendExample` selects first non-colliding root command and fills required values with generic placeholders. It does not know whether command is safe, interactive, destructive, or fully modeled.

Regression examples:

- #3136: `vault delete` with no path.
- #3141: interactive `packer console`.
- #3143: meaningless Ansible `"value"` host pattern.
- #3149: jq with no filter/input.
- #3150: Buildah add without operands.
- #3151: nonexistent Newman URL command.

Generic fix:

- Add shared example metadata to `CliToolDefinition`/`CliCommandDefinition`: safe-for-docs, interactive, destructive, required sample values, and preferred example command.
- Make common Markdown generator select only complete, safe, non-interactive commands.
- Omit runnable example when no qualified command exists.
- Validate chosen example against generated command model and argument renderer.
- Compile generated C# snippets and snapshot command/value selection for all tools.
- Use examples above as regressions, not per-tool Markdown patches.

Fix must propagate on regeneration to every generated CLI reference page.

thomhurst:
## Summary
- make runnable CLI documentation examples opt-in through shared command/tool safety metadata
- require preferred commands to be safe, non-interactive, non-destructive, and supplied with model-valid sample values
- omit runnable snippets when no qualified command exists instead of inventing generic placeholders
- curate safe Ansible, Buildah, jq, and Vault examples; explicitly reject the Packer/Newman and destructive regression cases
- compile every curated snippet against stubs generated from the same command model

## Validation
- guarded OptionsGenerator solution build (2 GB cap)
- guarded OptionsGenerator tests: 501 passed
- scoped `dotnet format whitespace`

Closes #3167

### DIFF
--- a/tools/ModularPipelines.OptionsGenerator/src/ModularPipelines.OptionsGenerator.Tests/Generators/MarkdownDocumentationGeneratorTests.cs
+++ b/tools/ModularPipelines.OptionsGenerator/src/ModularPipelines.OptionsGenerator.Tests/Generators/MarkdownDocumentationGeneratorTests.cs
@@ -254,6 +254,88 @@ public class MarkdownDocumentationGeneratorTests
     }
 
     [Test]
+    public async Task GenerateAsync_RejectsMissingPreferredCommand()
+    {
+        var tool = Tool("fake", Command("fake run", "FakeRunOptions", ["run"])) with
+        {
+            PreferredDocumentationExampleCommand = "fake status",
+        };
+
+        void GenerateDocumentation() =>
+            _ = new MarkdownDocumentationGenerator().GenerateAsync(tool);
+
+        await Assert.That(GenerateDocumentation)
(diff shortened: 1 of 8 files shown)


thomhurst:
Addressed the catalog review in e28df2977. All 46 registered CLIs now have an explicit documentation-example policy (curated or intentional omission), startup rejects unclassified registrations, and stale catalog/preferred command names fail loudly instead of silently dropping examples. I kept selection fail-closed rather than inferring safety from command verbs. OptionsGenerator tests pass 526/526. Release compilation completed, but the repository agent guard stopped the process tree after it exceeded the 2 GB local limit; not retried per repository policy.

PR #3217, opened 2026-07-27
Remove obsolete CI memory mitigations

### LINKED ISSUE #3212: CI: re-evaluate / remove the now-likely-vestigial build memory mitigations
## Background

While chasing the CI runner reclaim (#3179), several memory mitigations were added to `.github/workflows/dotnet.yml` under successive (and ultimately incorrect) theories:

- 10 GB **swap** step (`Increase swap space`)
- `MSBUILDDISABLENODEREUSE: "1"` (job env)
- `DOTNET_gcServer: "0"` (job env, workstation GC)

The **actual** root cause turned out to be a single test project (`GeneratedCliRenderingSnapshotTests`) referencing every tool integration at once, pulling the large AWS/Azure/Google SDK metadata into one `csc` compilation. That was fixed in PR #3201 by removing those snapshot tests, after which a **default-parallel build passes cleanly** (run [30249269627](https://github.com/thomhurst/ModularPipelines/actions/runs/30249269627) built + tested green with no serialization).

## Ask

These three levers were added before the real cause was understood and are likely no longer load-bearing. Evaluate removing them and confirm CI stays green:
- Drop the swap step and see if the build still fits (it should — peak memory is now bounded by normal per-project compiles).
- Drop `MSBUILDDISABLENODEREUSE` and `DOTNET_gcServer` overrides.

Remove one at a time (or together) so a red run pinpoints which, if any, is genuinely needed. Keep only what's actually required, with a one-line comment explaining why, so future readers don't treat them as unexplained ritual.

## References
- #3179, PR #3201

thomhurst:
## Summary
- remove the 10 GB Linux swap reconfiguration from the .NET workflow
- restore default server GC behavior
- retain `MSBUILDDISABLENODEREUSE=1` after the fully unmitigated run failed with `MSB4166`
- rely on the bounded per-project compilation shape established by #3201

## Test plan
- [x] Parse `.github/workflows/dotnet.yml` as YAML
- [x] Verify MSBuild node reuse remains disabled while server GC override and swap step are absent
- [x] Run `git diff --check`
- [x] Confirm the full GitHub Actions pipeline remains green with only node reuse disabled

Closes #3212

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -40,14 +40,6 @@ jobs:
     # still collected when this timeout trips.
     timeout-minutes: 60
     env:
-      # Tear down MSBuild worker/compiler nodes after each build instead of keeping them
-      # resident. BuildSolutionsModule builds several solutions in sequence; with node reuse
-      # the persistent nodes accumulate memory across those builds, which contributed to the
-      # runner reclaim in #3179. Disabling reuse keeps peak memory bounded per build.
-      MSBUILDDISABLENODEREUSE: "1"
-      # Workstation GC keeps the build/test footprint lower than server GC (which allocates a
-      # heap per core) on these multi-core runners.
-      DOTNET_gcServer: "0"
       # Single source of truth for the solutions CI restores and builds, so the Restore and
       # Build steps below can't drift apart. This mirrors BuildSolutionsModule.Solutions in
       # src/ModularPipelines.Build; keep the two in sync when adding a solution.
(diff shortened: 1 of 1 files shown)


thomhurst:
The first unmitigated Ubuntu run failed during `ModularPipelines.Examples.sln`: two MSBuild child nodes exited prematurely (`MSB4166`), then the build was canceled. That is consistent with memory pressure rather than a source failure.

Pushed e6a8ad783 as the first bisection step:

- restored only `MSBUILDDISABLENODEREUSE=1`
- kept server GC at the runtime default
- kept the 10 GB swap step removed
- rebased onto current `main`

Validation: workflow YAML parses; assertions confirm node reuse disabled, `DOTNET_gcServer` absent, and swap step absent. The new Ubuntu run will determine whether retained MSBuild nodes were the load-bearing mitigation.

PR #3357, opened 2026-07-28
Streamline coordinated console writes

### LINKED ISSUE #3255: CoordinatedTextWriter: per-character global locking and O(n^2) buffer scans on the redirected Console.Out
**Category:** perf — **Breaking:** no

`Console.Out` is replaced by `Console/CoordinatedTextWriter.cs`. `Write(char)` → `Write(value.ToString())` (`:96-99`) takes the global `_lineBufferLock` per character; each write calls `GetSecretPatterns()` (LINQ `Distinct().OrderByDescending().ToArray()` over all secrets, `:136`), `state.Buffer.ToString()` (full copy), and `FlushSafePrefix`'s `ToString(0, safeLength).IndexOf('\n')` loop (`:165-170`) — O(n^2) in accumulated buffer length. Any library writing char-by-char degrades to quadratic with global serialisation across all modules.

**Fix:** cache the pattern array; real `Write(char)`/`Write(char[],int,int)` overrides appending directly; scan the `StringBuilder` for newlines without `ToString()`.

Part of #3239.

thomhurst:
## Summary
- cache coordinated-writer secret patterns by provider version
- append `char`, `char[]`, and span writes without intermediate strings
- defer buffered partial-line processing until needed
- scan `StringBuilder` line data once and retain only bounded secret-prefix data

## Validation
- `SecretMaskingPatternTests`: 23 passed
- `ModularPipelines.sln` Release build: 0 errors (250 existing warnings)

## Coordination
#3356 introduces the same provider-version invalidation foundation for `SecretObfuscator`; this branch will rebase after that PR merges so this diff remains focused.

Closes #3255

### DIFF
--- a/src/ModularPipelines/Console/CoordinatedTextWriter.cs
+++ b/src/ModularPipelines/Console/CoordinatedTextWriter.cs
@@ -1,3 +1,4 @@
+using System.Buffers;
 using System.Diagnostics.CodeAnalysis;
 using System.Text;
 using ModularPipelines.Engine;
@@ -36,6 +37,8 @@ internal class CoordinatedTextWriter : TextWriter
     private readonly ISecretProvider _secretProvider;
     private readonly Dictionary<LineBufferKey, LineBufferState> _lineBuffers = [];
     private readonly object _lineBufferLock = new();
+    private string[] _secretPatterns = [];
+    private long? _secretPatternsVersion;
 
     /// <summary>
     /// Initialises a new instance of the <see cref="CoordinatedTextWriter"/> class.
@@ -68,7 +71,7 @@ internal class CoordinatedTextWriter : TextWriter
     {
         lock (_lineBufferLock)
         {
-            WriteCore(value ?? string.Empty, appendNewLine: true);
+            WriteCore((value ?? string.Empty).AsSpan(), appendNewLine: true);
         }
     }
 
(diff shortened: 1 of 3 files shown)


thomhurst:
Rebased onto current \main\ and resolved the provider-version overlap by using the merged atomic \SecretSnapshot\ foundation. The writer retains a stable-version fast path, so repeated character writes do not allocate snapshots. Validation: 23 \SecretMaskingPatternTests\ passed; core Release build has 0 errors. Remote SHA verified at \d5a9d9999\.

thomhurst:
Fixed the Ubuntu pipeline hang in 95800d06f. The optimized snapshot cache dereferenced `Secrets` from a default `SecretSnapshot`; the resulting exception during direct logger replay recursively re-entered output handling and wedged the flush queue. The snapshot reader now treats a default/null secret list as empty.

Validation: reproduced the focused test hang locally, then verified it passes in 379 ms; all 16 `OutputCoordinatorTests` and 23 `SecretMaskingPatternTests` pass; core Release build succeeds with 0 errors (250 existing warnings).

thomhurst:
Fixed the latest Ubuntu failure in 5b35563f2 after rebasing onto current \main\. The failing test awaited both buffer completions, but those complete before the queue's \inally\ awaits \ResumeAsync\; its mock verification therefore raced the processor. The test now awaits \WaitForPendingFlushesAsync\ before verifying pause/resume. All 17 \OutputCoordinatorTests\ pass; unit-test Release build succeeds with 0 errors.

PR #3358, opened 2026-07-28
Make scheduler event-driven

### LINKED ISSUE #3256: Scheduler loop: make event-driven; O(n^2) allocations per 100ms cycle under the exclusive write lock
**Category:** perf — **Breaking:** no

Two compounding problems:
1. `Engine/ModuleScheduler.cs:420-438,455-477` + `SchedulerOptions.NotificationTimeout = 100ms` — the loop wakes ≥10×/s, takes the exclusive write lock twice per iteration (`FindReadyModules`, `ShouldExitScheduler`) and does two full O(n) scans of `_moduleStates`. A 30-minute pipeline ≈ 18,000 exclusive-lock acquisitions contending with every `MarkModuleStarted`/`MarkModuleCompleted`.
2. `ModuleScheduler.cs:655` — a fresh list of all active modules is materialised per ready module inside the write lock, and `Scheduling/ModuleConstraintEvaluator.cs:21,98` does `.ToList()` twice more plus `RequiredLockKeys.Intersect(...).ToArray()` per pair.

**Fix:** purely event-driven loop (the notification semaphore exists; keep the timeout only for deadlock detection); incrementally-maintained counters; build the active list once per cycle; short-circuit constraint evaluation when no module declares lock keys/sequential execution.

Part of #3239.

thomhurst:
Closes #3256

## Summary
- wait on scheduler notifications during normal execution and reserve the timeout for deadlock confirmation
- maintain module-state counters incrementally instead of rescanning all states for exit checks
- build the active-module list once per constrained cycle and skip constraint evaluation entirely for unconstrained pipelines
- remove repeated constraint evaluator list/intersection allocations and restore the deferred-module reschedule notification

## Validation
- `ModuleScheduler*Tests`: 16/16 passed
- `TaskCompletionSourceTests`: 5/5 passed
- `ModuleExecutorLoggingTests`: 6/6 passed
- `ModularPipelines.sln` Release build: 0 errors (251 existing warnings)

Full solution formatting verification remains blocked by pre-existing repository-wide violations in untouched files.

### DIFF
--- a/src/ModularPipelines/Engine/ModuleScheduler.cs
+++ b/src/ModularPipelines/Engine/ModuleScheduler.cs
@@ -32,6 +32,7 @@ internal class ModuleScheduler : IModuleScheduler
     private readonly HashSet<ModuleState> _queuedModules;
     private readonly HashSet<ModuleState> _executingModules;
     private readonly ModuleStateQueries _stateQueries;
+    private readonly ModuleStateCounters _stateCounters;
     private readonly SchedulerExitConditions _exitConditions;
     private readonly Channel<ModuleState> _readyChannel;
     private readonly SemaphoreSlim _schedulerNotification;
@@ -39,6 +40,7 @@ internal class ModuleScheduler : IModuleScheduler
     private readonly ReaderWriterLockSlim _stateLock = new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
     private readonly IModuleStateTracker _stateTracker;
 
+    private bool _hasSchedulingConstraints;
     private bool _schedulerCompleted;
     private int _disposeState;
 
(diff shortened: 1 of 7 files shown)


thomhurst:
Review follow-up in 4785e9aa7: removed the now-unreferenced O(n) ModuleStateSnapshot.Create factory and corrected NotificationTimeout docs to describe deadlock confirmation. I also audited the heartbeat concern: SchedulerStatusReporter suppresses an unchanged SchedulerStatusSnapshot, so the previous 100ms loop did not emit a repeating 15-second heartbeat; it only re-invoked a method that returned without logging. Adding a timer would therefore introduce new logging behavior, not preserve existing behavior, so the event-driven indefinite wait remains. Validation: ModuleSchedulerDynamicCycleTests 13/13; core Release build 0 errors.

PR #3359, opened 2026-07-28
Consolidate progress rendering work

### LINKED ISSUE #3258: ProgressSession: per-module ticker tasks, uncancellable delays, fire-and-forget render loop
**Category:** perf + robustness — **Breaking:** no

`Console/ProgressSession.cs`:
- `:411-440` — `StartProgressTicker` spawns one detached `Task.Run` + 1s `Task.Delay` loop per module and per sub-module, each taking the shared `_progressLock` the render loop also holds; the tasks use `CancellationToken.None` so they survive session disposal. 200 modules = 200 timer tasks.
- `:481` — `PauseAsync` ends with unconditional `await Task.Delay(50)`, invoked per buffer flush inside `OutputCoordinator`'s `_writeLock` (`OutputCoordinator.cs:308`); with 40 in-flight modules each live-flush tick costs ≥2s of pure sleeping while holding the write lock.
- `:74-80,178` — `_ = RunProgressLoopAsync()` fire-and-forget, delays ignore the session token; `:542` — `DisposeAsync` creates a 5s `Task.Delay` never cancelled (timer leak).

**Fix:** one ticker iterating the live task set driven by the existing 100ms refresh loop; pause once around a flush batch with a real "refresh drained" signal; store/await the loop task and flow tokens.

Part of #3239.

thomhurst:
## Summary
- update module progress estimates from the existing 100 ms render loop instead of per-module tasks
- own, cancel, and await the progress loop during disposal
- pause progress once while draining each queued output batch

## Validation
- `OutputCoordinatorTests`: 17 passed
- `ProgressSessionTests`: 1 passed
- `ModularPipelines.sln` Release build: 0 errors

Closes #3258

### DIFF
--- a/src/ModularPipelines/Console/OutputCoordinator.cs
+++ b/src/ModularPipelines/Console/OutputCoordinator.cs
@@ -232,11 +232,27 @@ internal sealed class OutputCoordinator : IOutputCoordinator
         try
         {
             var formatter = _formatterProvider.GetFormatter();
-            PendingFlush? pending;
-
-            while ((pending = DequeuePendingFlush()) is not null)
+            await _writeLock.WaitAsync().ConfigureAwait(false);
+            try
             {
-                await ProcessPendingFlushAsync(pending, formatter).ConfigureAwait(false);
+                var progressController = _progressController;
+                await progressController.PauseAsync().ConfigureAwait(false);
+                try
+                {
+                    PendingFlush? pending;
+                    while ((pending = DequeuePendingFlush()) is not null)
+                    {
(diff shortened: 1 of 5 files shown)


thomhurst:
Addressed the PauseAsync reentrancy note in a3e83c599: repeated pause calls now complete immediately instead of waiting on a potentially stale refresh completion source; the first caller still owns any active-refresh wait. Validation: ProgressSessionTests 1/1, OutputCoordinatorTests 17/17, core Release build 0 errors.

PR #3360, opened 2026-07-28
Cache module attribute metadata

### LINKED ISSUE #3260: Cache module attribute reflection across the engine (repeated per event, per call, with instance-identity bugs)
**Category:** perf + correctness — **Breaking:** no

Three uncached attribute-reflection hot paths:
1. `Engine/PipelineSetupExecutor.cs:74` — `Attribute.GetCustomAttributes(moduleType).ToList().AsReadOnly()` on every `OnModuleReady/Start/End/Failure/Skipped` (3-4× per module), even with zero hooks registered.
2. `Engine/Attributes/ModuleAttributeEventService.cs:18-19` — `GetAttributes` is the only uncached accessor, called per module execution (`ModuleRunner.cs:210`, `RegistrationEventExecutor.cs:88`). It also returns *different attribute instances* than the cached handler lists, so an attribute holding state between `OnStart`/`OnEnd` sees two different objects — a real correctness hazard.
3. `Engine/Dependencies/ModuleMetadataRegistry.cs:70-128` — `FinalizeMetadata` re-runs attribute reflection ~5-6× per module at startup with no idempotency guard; `HasAttribute`/`GetAttribute(s)` are uncached and called O(candidates × predicate attributes) in `ModuleDependencyResolver`.

**Fix (tactical):** memoise per type; return identical instances. **Fix (strategic, v4):** source-generate a `GeneratedModuleMetadata` table (pattern exists at `Engine/Attributes/GeneratedModuleEventMetadata.cs`) and delete the runtime reflection — complements #3226/#3227.

Part of #3239.

thomhurst:
Closes #3260

## Summary
- cache each module type's attribute instances once, including concurrent access
- reuse that cache across event handlers, hook contexts, registration queries, and metadata
- skip module-hook context/attribute creation when no module hooks are registered
- memoize finalized metadata while preserving reflection ambiguity semantics

## Validation
- targeted unit tests: 34 passed
- metadata cross-phase integration: 1 passed
- flexible dependency integration: 15 passed
- `ModularPipelines.sln` Release build: 0 errors (250 existing warnings)

### DIFF
--- a/src/ModularPipelines/Context/ModuleRegistrationContext.cs
+++ b/src/ModularPipelines/Context/ModuleRegistrationContext.cs
@@ -1,5 +1,4 @@
 // src/ModularPipelines/Context/ModuleRegistrationContext.cs
-using System.Reflection;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -82,7 +81,7 @@ internal class ModuleRegistrationContext : IModuleRegistrationContext
         => _registeredModuleTypes.Where(t => typeof(TBase).IsAssignableFrom(t) && t != ModuleType);
 
     public IEnumerable<Type> GetModulesWithAttribute<TAttribute>() where TAttribute : Attribute
-        => _registeredModuleTypes.Where(t => t.GetCustomAttribute<TAttribute>() != null);
+        => _registeredModuleTypes.Where(_metadataRegistry.HasAttribute<TAttribute>);
 
     public void AddDependency<TModule>() where TModule : IModule
         => AddDependency(typeof(TModule));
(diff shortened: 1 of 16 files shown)


thomhurst:
Addressed both review suggestions in f28e05adc. Removed the one-argument ModuleMetadataRegistry constructor so every construction path must provide the shared IModuleAttributeEventService explicitly; updated unit/distributed fixtures. Added interface remarks requiring FinalizeMetadata after registration tag/category mutation before reads. Validation: 8/8 metadata tests, 9/9 distributed publisher tests, core Release build 0 errors, focused formatting.

thomhurst:
Fixed the parallel-test race in 4fcad5248 by serializing `PipelineSetupExecutorTests` with a class-level TUnit `NotInParallel` constraint. Focused class: 2/2 passed; scoped whitespace formatting and diff checks passed; rebased onto current `main`.

PR #3369, opened 2026-07-28
Initialize module configuration once

### LINKED ISSUE #3281: Module.Configuration race and Tags allocating per access
**Category:** bug + perf — **Breaking:** no

`Modules/Module.cs:95` — `_configuration ??=` is not thread-safe; `IModule.Configuration` is read concurrently from scheduler threads, so user `Configure()` may run twice and two different `ModuleConfiguration` instances can be observed. `Tags` (`Module.cs:109`) allocates a new `HashSet` on every access.

**Fix:** compute configuration once during activation (`IModuleActivator`) and store; cache `Tags` as a `FrozenSet`.

Part of #3239.

thomhurst:
## Summary
- initialize module configuration once during activation under the module logging context
- use thread-safe lazy initialization for directly constructed or pre-created modules
- cache programmatic module tags as a case-insensitive FrozenSet

## Validation
- `ModularPipelines.sln` Release build (0 errors)
- `ModuleConfigureTests` (6 passed)
- `ModuleMetadataRegistryTests` (5 passed)
- `FlexibleDependencyIntegrationTests` (15 passed)

Closes #3281

### DIFF
--- a/src/ModularPipelines/Engine/ModuleActivator.cs
+++ b/src/ModularPipelines/Engine/ModuleActivator.cs
@@ -27,7 +27,9 @@ internal sealed class ModuleActivator : IModuleActivator
 
         try
         {
-            return (IModule)ActivatorUtilities.CreateInstance(serviceProvider, moduleType);
+            var module = (IModule) ActivatorUtilities.CreateInstance(serviceProvider, moduleType);
+            _ = module.Configuration;
+            return module;
         }
         finally
         {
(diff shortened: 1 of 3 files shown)


thomhurst:
Applied the reviewer simplification in b5b2e70c5: ModuleActivator now eagerly reads the existing IModule.Configuration property, removing IModuleConfigurationInitializer and its runtime type check. ModuleConfigureTests pass 6/6 on the rebuilt final state.

thomhurst:
Reviewed the later Tags access-path note. I am not renaming/removing the public virtual Tags override hook here because #3281 explicitly requires a non-breaking fix; that change would break consumer overrides. Current framework reads use the cached ITaggedModule path, the default public property no longer allocates, and the reviewer confirmed no internal concrete Tags reads. Treating the broader API redesign as out of scope for this bug fix.

PR #3397, opened 2026-07-29
Dispose engine cancellation plumbing

### LINKED ISSUE #3248: Cancellation plumbing leaks: undisposed linked CTS per module, discarded CancellationTokenRegistration, never-unsubscribed global handlers
**Category:** engine / bug — **Breaking:** no

Three leak sites:
1. `Engine/ModuleExecutionPipeline.cs:209-211` — the linked `CancellationTokenSource` per module is never disposed (nor is the source it replaces); each adds a permanent callback node to the singleton engine token. Unbounded with matrix/sub-module expansion.
2. `Engine/Executors/ExecutionOrchestrator.cs:108` — `cancellationToken.Register(...)` return value discarded; registration outlives the pipeline and roots the orchestrator.
3. `Engine/EngineCancellationToken.cs:53-59` — subscribes `Console.CancelKeyPress` + `AppDomain.ProcessExit` in the ctor, never unsubscribes. Hosts building multiple pipelines (tests; `ValidateAsync` then `BuildAsync`) accumulate handlers that call `Cancel()` on disposed instances.

**Fix:** dispose linked CTS in the module runner's `finally`; capture and dispose registrations; unsubscribe in `Dispose`.

Part of #3239.

thomhurst:
## Summary
- dispose per-module linked cancellation token sources on every exit path
- scope caller-token registrations to orchestration lifetime
- detach process-wide cancellation event handlers when the engine token is disposed
- add regression coverage for disposal and handler rooting

## Validation
- targeted ModuleExecutionPipelineTests: 1/1 passed
- targeted ExecutionOrchestratorTests: 2/2 passed
- targeted EngineCancellationTokenTests: 6/6 passed
- core ModularPipelines.sln Release build: passed
- whitespace verification: passed
- full unit-test project exceeded the required 2 GB agent guard; not retried with higher limits

Closes #3248

### DIFF
--- a/src/ModularPipelines/Engine/EngineCancellationToken.cs
+++ b/src/ModularPipelines/Engine/EngineCancellationToken.cs
@@ -22,7 +22,7 @@ internal class EngineCancellationToken : IDisposable
     public CancellationToken Token => _cts.Token;
 
     /// <summary>
-    /// Gets whether cancellation has been requested on the underlying token source.
+    /// Gets a value indicating whether cancellation has been requested on the underlying token source.
     /// </summary>
     public bool IsCancellationRequested => _cts.IsCancellationRequested;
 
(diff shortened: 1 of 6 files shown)


thomhurst:
Reviewed the timeout/disposal question. On the timed path the module receives timeoutCts.Token from TimeoutHelper, and TimeoutHelper already disposes that source when it returns after the grace period; disposing the outer module CTS here therefore does not introduce the cited timeout-path lifetime. Deferring cancellation-source disposal until an abandoned execution task completes would need to be solved in TimeoutHelper as a separate ownership change. Keeping this PR scoped to the leaks in #3248.

Issue #3307, opened 2026-07-28
Add source-generator snapshot and incremental-cache tests


thomhurst:
**Category:** test coverage — **Breaking:** no

`src/ModularPipelines.SourceGenerator/` has no test project at all; nothing outside the project references the generator types. Regressions (like the broken incremental caching) are invisible.

**Fix:** `Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing` project with snapshot tests plus explicit incremental-cache-hit assertions (`GeneratorDriver` step reason `Cached`).

Part of #3239.

PR #3412, opened 2026-07-29
Reject generated module accessor collisions

### LINKED ISSUE #3305: SourceGenerator: duplicate module class names silently collapse into one extension method
**Category:** source generator / bug — **Breaking:** no

`ModuleExtensionsGenerator.cs:174-178` — `modules.GroupBy(m => m.ClassName)` then `.First()`: two `BuildModule` classes in different namespaces yield one `GetBuildModule()` bound to an arbitrary one. Generated code also uses short type names with bulk `using` imports (`:158-161`), so it can emit ambiguous-reference compile errors with no explanation.

**Fix:** emit `global::`-qualified names; report a diagnostic on collision (or suffix by namespace).

Part of #3239.

thomhurst:
## Summary
- emit fully qualified module type names in generated accessors
- report `MPGEN002` when multiple module types map to the same accessor name
- skip collided accessors instead of selecting an arbitrary module
- cover duplicate names and qualified output with generator tests

## Validation
- Source generator tests: 6 passed
- Incremental generator caching tests: 3 passed
- `ModularPipelines.sln` Release build: 0 errors (existing warnings remain)

Closes #3305


thomhurst:
Addressed the diagnostic-location review in df39c086ce. ModuleClassInfo now carries the stable first symbol declaration location, and MPGEN002 is reported at the first conflicting module instead of Location.None. The collision test now asserts an in-source location at the expected declaration. Source-generator tests pass 6/6; focused project formatting is warning-clean.

thomhurst:
Fixed the Ubuntu pipeline failure in cdc8b43e12. The source-bound Location added for MPGEN002 had become part of ModuleClassInfo equality, invalidating generated output after trivia-only edits (and defeating partial-type deduplication). Generation now consumes stable module identity while a separate location-bearing pipeline emits collision diagnostics. Local repro changed from failing to passing; source-generator tests pass 7/7 and the Release development-analyzer suite passes 29/29.

thomhurst:
CI root cause is resolved in cdc8b43e12: location-bearing diagnostic candidates are now separated from stable generation metadata, restoring trivia-only incremental caching while preserving navigable MPGEN002 locations.

The remaining shared-test-harness suggestion is intentionally deferred to #3307, which is the queued source-generator test-infrastructure work. Extracting the sibling test class in this collision-fix PR would broaden scope without changing production behavior.

Latest source-generator tests are 7/7 and the current CI cycle has no failures; remaining checks are pending.

PR #3413, opened 2026-07-29
Require complete CommandResult data

### LINKED ISSUE #3278: v4: CommandResult — required members instead of CS8618 suppression
**Category:** API reshape — **Breaking:** yes

`Models/CommandResult.cs:44-51` — `#pragma warning disable CS8618` around `internal CommandResult(Command command)` leaves `StandardOutput`/`StandardError`/`ExitCode`/timings unset while the public properties are declared non-nullable.

**Fix:** `record` with `required` members (or split `CommandResult` / `PendingCommandResult`); delete the suppression.

Part of #3239.

thomhurst:
## Summary
- make every `CommandResult` data property a required init-only member
- annotate constructors that fully initialize the required contract
- remove the CS8618 suppression
- return valid empty output and zero-duration timing data for dry runs
- cover required metadata and dry-run invariants

## Validation
- CommandResult tests: 2 passed
- CommandException tests: 1 passed
- CommandBuilderBase tests: 23 passed
- `ModularPipelines.sln` Release build: 0 errors (existing warnings remain)

## Breaking change
`CommandResult` properties now expose a required init-only contract.

Closes #3278

### DIFF
--- a/src/ModularPipelines/Models/CommandResult.cs
+++ b/src/ModularPipelines/Models/CommandResult.cs
@@ -9,12 +9,13 @@ public record CommandResult
     /// <summary>
     /// Gets the command that was executed.
     /// </summary>
-    public string CommandInput { get; }
+    public required string CommandInput { get; init; }
 
-    public IReadOnlyDictionary<string, string?> EnvironmentVariables { get; }
+    public required IReadOnlyDictionary<string, string?> EnvironmentVariables { get; init; }
 
-    public string WorkingDirectory { get; }
+    public required string WorkingDirectory { get; init; }
 
+    [SetsRequiredMembers]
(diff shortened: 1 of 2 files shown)


thomhurst:
Reviewed the two non-blocking observations:

- Keeping the internal constructors explicit. A helper cannot assign these `init` properties outside a constructor, while chaining the dry-run constructor would need extra factory/tuple machinery to capture one timestamp for both `StartTime` and `EndTime`. Repeating the three direct `Command` mappings is smaller and clearer.
- Dry-run result output intentionally remains empty; the existing placeholder log text is presentation-only and predates this required-data contract.

No code change needed. The current head has 2/2 focused tests and green CI.

PR #3560, opened 2026-07-30
Write CI workflow commands directly to stdout

### LINKED ISSUE #3549: Security: CI workflow commands (::add-mask::) are routed through Spectre rendering and can leak the mask pattern into the log
**Category:** secret hygiene — **Breaking:** no

**Rotate the affected credential before/alongside fixing** — the pattern text (containing a base64 form of a registered secret) has appeared in public run logs on multiple recent main runs.

`BuildSystemSecretMasker` (`Engine/BuildSystemSecretMasker.cs:70`) emits `::add-mask::<pattern>` — where `SecretMaskingPatternGenerator.Generate` (`Engine/SecretMaskingPatternGenerator.cs:18`) includes base64 variants of registered secrets — via `ConsoleWriter.LogToConsole` -> Spectre `AnsiConsole.MarkupLine` (`ConsoleWriter.cs:14`) instead of an atomic raw `stdout.WriteLine`. Workflow commands must reach the runner as a single unmodified physical line; routed through the Spectre pipeline they can be split/mangled, in which case the runner never registers the mask AND the pattern text itself (recoverable secret material) lands in the log. Observed deterministically at pipeline start on the three newest successful main runs.

**Fix:** write all CI workflow commands (`::add-mask::`, `::group::`, `::notice::`, `##vso[...]` etc.) directly to the original unredirected stdout with a single plain `WriteLine`, bypassing Spectre rendering and the intercepting/coordinating writers. Add a CI assertion that no generated mask-pattern text appears in the job log.

Part of #3239 (round 2 addendum).

thomhurst:
## Summary

- add a dedicated raw command writer that captures original stdout and emits each CI command as one physical line
- route secret masks, CI issue annotations, and CI group boundaries around Spectre/intercepting writers while preserving rich local headers
- escape Azure secret-mask command values and add regressions proving generated mask patterns never enter rich-console output

## Validation

- `ModularPipelines.UnitTests`: 46 focused security/output tests passed
- `ModularPipelines.sln` Release build: 0 errors; 217 existing warnings
- touched-file whitespace verification passed
- full core test run exceeded the mandated 2 GB agent guard (2,828 MB); CI owns full-suite validation
- solution-wide whitespace verification still reports nine unrelated baseline violations in untouched HTTP/timeout files

Credential rotation for any values exposed by historical CI logs remains an external owner/security action.

Closes #3549

### DIFF
--- a/src/ModularPipelines.Azure.Pipelines/AzurePipeline.cs
+++ b/src/ModularPipelines.Azure.Pipelines/AzurePipeline.cs
@@ -50,19 +50,13 @@ internal class AzurePipeline : IAzurePipeline
             _formatter = formatter;
 
             var startCommand = formatter.GetStartBlockCommand(name);
-            if (startCommand != null)
-            {
-                _buffer.WriteLine(startCommand);
-            }
+            buffer.WriteGroupCommand(formatter, startCommand);
         }
 
         public void Dispose()
         {
             var endCommand = _formatter.GetEndBlockCommand(_name);
-            if (endCommand != null)
-            {
-                _buffer.WriteLine(endCommand);
-            }
+            _buffer.WriteGroupCommand(_formatter, endCommand);
         }
     }
 }
--- a/src/ModularPipelines.GitHub/GitHub.cs
+++ b/src/ModularPipelines.GitHub/GitHub.cs
@@ -85,19 +85,13 @@ internal class GitHub : IGitHub
             _formatter = formatter;
 
             var startCommand = formatter.GetStartBlockCommand(name);
-            if (startCommand != null)
-            {
-                _github._buffer.WriteLine(startCommand);
-            }
+            github._buffer.WriteGroupCommand(formatter, startCommand);
         }
 
         public void Dispose()
         {
             var endCommand = _formatter.GetEndBlockCommand(_name);
-            if (endCommand != null)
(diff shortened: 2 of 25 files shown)


thomhurst:
Addressed the DI-order concern in `25be54c61a`.

`BuildSystemCommandWriter` is now constructed inside `RegisterBuildSystemServices` from the then-current `Console.Out` and registered as an instance. The parameterless lazy-capture constructor is gone, so resolving the service after `ConsoleCoordinator.Install()` cannot capture the intercepted writer.

Added a nonparallel regression that registers services against one writer, redirects `Console.Out` before provider construction/service resolution, then proves the workflow command still reaches only the registration-time writer. `OutputCoordinator` remains separately captured, but its construction is a direct `ConsoleCoordinator` constructor prerequisite, so `Install()` cannot precede that capture.

Validation: writer tests 3/3, dependency-injection tests 3/3, changed-file analyzer verification clean, Release core build 0 errors (227 existing warnings).

@codex review

thomhurst:
Rebased onto current `main` at `8c247942fb`. Patch equivalence preserved (`726e98f15bf532ff4dcef46e1929aef979c26a0a` before and after). New head: `1f5d5f634b`.

@codex review

PR #3625, opened 2026-08-02
Make asynchronous tests deterministic

### LINKED ISSUE #3522: Tests: Task.Delay-then-assert races in logger/output and distributed executor tests
**Category:** test reliability — **Breaking:** no

`DeferredCommandOutputLoggerTests.cs:115` — `Task.Delay(50)` then `AttemptCount == 1` against a 10ms debounce timer (fails if the timer has not fired in 50ms, and if it fires twice). `ModuleLoggerTests.cs:194-199` — `Task.Delay(50)` then `disposeTask.IsCompleted IsFalse`. `DistributedModuleExecutorTests.cs:696-700,731-735,837-852` — fire-and-forget `_ = Task.Run(async () => { await Task.Delay(...); Publish...})` where a publish exception is silently unobserved and surfaces as a hang caught only by HangDump.

**Fix:** replace fixed delays with TCS handshakes or the existing bounded `WaitUntilAsync` helper pattern (`CommandLoggerTests.cs:473-487`); await the background publish tasks at test end so failures are observed. Also shaves ~1-2s of pure sleep off the distributed suite.

Part of #3239 (round 2).

thomhurst:
## Summary
- replace fixed-delay logger assertions with completion and admission handshakes
- publish distributed results only after result waits begin
- synchronize worker registration with the worker-query barrier

## Validation
- DeferredCommandOutputLoggerTests: 6/6 passed
- ModuleLoggerTests/Dispose_WaitsForInProgressLogAdmission: 1/1 passed
- DistributedModuleExecutorTests: 27/27 passed
- ModularPipelines.UnitTests Release build: 0 warnings, 0 errors
- ModularPipelines.Distributed.UnitTests Release build: 0 warnings, 0 errors

Fixes #3522

### DIFF
--- a/src/ModularPipelines.Analyzers/ModularPipelines.Analyzers.Test/ModularPipelinesAnalyzersUnitTests.cs
+++ b/src/ModularPipelines.Analyzers/ModularPipelines.Analyzers.Test/ModularPipelinesAnalyzersUnitTests.cs
@@ -292,9 +292,11 @@ public class Module2 : Module<string>
             using ModularPipelines.Attributes;
 
             namespace Example;
+
             public class Dependency : Module<string>
             {
-                protected override Task<string?> ExecuteAsync(IModuleContext context, CancellationToken cancellationToken) => Task.FromResult<string?>(null);
+                protected override Task<string?> ExecuteAsync(IModuleContext context, CancellationToken cancellationToken)
+                    => Task.FromResult<string?>(null);
             }
 
             [DependsOn<Dependency>]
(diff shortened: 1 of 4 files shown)


thomhurst:
Pushed 4d5111a99f to address both findings. The dispose test now asserts the dispose task remains blocked before releasing the in-flight log, preserving the behavior named by the test. I also aligned the inherited analyzer code-fix snapshot with formatter output, which fixes the Ubuntu failure in RunConsumerAnalyzersUnitTestsModule. Focused Release validations pass for both tests (1/1 each).

PR #3636, opened 2026-08-02
Replace skipped FTP tests with loopback server

### LINKED ISSUE #3529: Tests: FTP integration has zero executing tests - entire class permanently skipped against a public server
**Category:** test coverage — **Breaking:** no

`test/ModularPipelines.Ftp.UnitTests/Helpers/FtpTests.cs:11` — class-level `[Skip("FTP tests flaky due to server load")]`; the tests target ftp.pureftpd.org (20, 43). With the class skipped and no smoke-props import in the csproj, ModularPipelines.Ftp ships with no executing tests at all.

**Fix:** replicate the #3334 SMTP fix — an in-proc FTP stub (TcpListener, port 0) covering GetFtpClientAsync/download/dispose semantics — then delete the Skip. At minimum import the smoke props so the package is not entirely untested.

Part of #3239 (round 2).

thomhurst:
Closes #3529.

## Summary
- replace the permanently skipped public-server tests with a loopback FTP stub
- exercise FluentFTP `AutoConnect`, passive-mode download, and response content
- verify `IFtp` disposes tracked clients
- remove external network/server-load dependence and clean temporary output

## Validation
- focused FtpTests: 2/2 passed
- FTP integration Release build: 0 warnings, 0 errors
- test-project whitespace verification passed

### DIFF
--- a/test/ModularPipelines.Ftp.UnitTests/Helpers/FtpTests.cs
+++ b/test/ModularPipelines.Ftp.UnitTests/Helpers/FtpTests.cs
@@ -8,45 +8,70 @@ using File = ModularPipelines.FileSystem.File;
 
 namespace ModularPipelines.Ftp.UnitTests.Helpers;
 
-[Skip("FTP tests flaky due to server load")]
 public class FtpTests : TestBase
 {
     [Test]
-    [NotInParallel(nameof(FtpTests), Order = 1)]
     public async Task Can_Download()
     {
+        await using var ftpServer = LocalFtpServer.Start();
         var ftp = await GetService<IFtp>();
 
-        var client = await ftp.GetFtpClientAsync(new FtpOptions("ftp.pureftpd.org", new NetworkCredential())
-        {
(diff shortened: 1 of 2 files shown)


thomhurst:
Ubuntu exposed a deterministic fixture lifetime bug: LocalFtpServer cancelled itself after 15 seconds, while FluentFTP AutoDetect can still be probing until roughly 24 seconds on Linux. Commit 035aed7293 removes that artificial timeout so async disposal owns cancellation. Focused FTP tests pass 2/2, targeted formatting passes, and the FTP Release solution build passes with 0 warnings/errors.

thomhurst:
Addressed the remaining useful review notes in \c7a6def7e\: passive data listeners now belong to a per-control-connection \FtpSession\, removing cross-client state races; download cleanup now uses \TempFile\. FTP tests pass 2/2 and the FTP Release solution builds with 0 warnings/errors. Targeted format completed.

PR #3637, opened 2026-08-02
Add non-generic Module base class

### LINKED ISSUE #3492: Reinstate a non-generic Module base class for fire-and-forget modules
**Category:** DX — **Breaking:** no (additive)

`docs/docs/how-to/defining-modules.md:25-56` now teaches `Module<None>` with `return None.Value;` for the extremely common cleanup/notify/publish case (`Models/None.cs`). Most CI modules return nothing — the template's own Restore/Build/Test/Publish return `CommandResult` only because the tool call happens to produce one. Cake/NUKE targets are void; here the simplest possible module carries a generic parameter, a nullable task type, and a sentinel return.

**Fix:** reinstate `public abstract class Module : Module<None>` with `protected abstract Task ExecuteAsync(IModuleContext, CancellationToken)` (sealed adapter returning `None.Value`), mirroring the `SyncModule<T>` adapter pattern (`Modules/SyncModule.cs:72-75`). Name the abstract method `ExecuteAsync` to stay consistent with #3287. Update README/template/docs to use it for void modules.

Part of #3239 (round 2).

thomhurst:
Closes #3492.

## Summary

- add a non-generic `Module` base for modules that return no meaningful value
- expose `protected abstract Task ExecuteAsync(IModuleContext, CancellationToken)` to module authors
- seal the infrastructure adapter that awaits the operation and returns `None.Value` to the existing `Module<None>` engine path
- migrate README, compiled snippets, and current how-to docs away from `Module<None>` for asynchronous void modules
- retain `SyncModule<None>` for synchronous void modules

## Design

C# cannot overload methods by return type alone, so a hidden `NonGenericModuleAdapter : Module<None>` owns the sealed `Task<None>` bridge while `Module` exposes the user-facing `Task ExecuteAsync` override. Generic `Module<T>` keeps its abstract compile-time contract unchanged.

## Validation

- `ModularPipelines.sln` Release build: passed, 0 warnings/errors
- `ModularPipelines.UnitTests` Release build: passed, 0 warnings/errors
- `NonGenericModuleTests`: 1/1 passed
- `ExecutionApiSurfaceTests`: 1/1 passed
- `DocumentationSnippetTests`: 4/4 passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,14 +86,13 @@ public class BuildModule : Module<BuildInfo>
 
 // PublishModule retrieves and uses it
 [DependsOn<BuildModule>]
-public class PublishModule : Module<None>
+public class PublishModule : Module
 {
-    protected override async Task<None> ExecuteAsync(IModuleContext context, CancellationToken cancellationToken)
+    protected override async Task ExecuteAsync(IModuleContext context, CancellationToken cancellationToken)
     {
         var buildResult = await context.GetModule<BuildModule>();
         var outputPath = buildResult.Value.OutputPath; // Throws with module context if unavailable
         // Publish using the build output...
-        return None.Value;
     }
 }
 ```
--- a/README_Template.md
+++ b/README_Template.md
@@ -86,14 +86,13 @@ public class BuildModule : Module<BuildInfo>
 
 // PublishModule retrieves and uses it
 [DependsOn<BuildModule>]
-public class PublishModule : Module<None>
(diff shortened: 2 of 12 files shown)


thomhurst:
Merged current main after the core compile repair. Core Release build passes with 0 warnings/errors; NonGenericModuleTests and ExecutionApiSurfaceTests pass (1/1 each). Review audit: 0 unresolved threads.

PR #3640, opened 2026-08-02
Add typed module registration handles

### LINKED ISSUE #3494: v4: builder WithTags/WithCategory rely on hidden LastRegisteredModuleType state and fail at runtime in legal call orders
**Category:** api-reshaping — **Breaking:** yes

`Extensions/PipelineBuilderExtensions.cs:120-138` configure "the most recently registered module" via `builder.LastRegisteredModuleType` (`PipelineBuilder.cs:41`); `AddModulesFromAssembly` sets it to null (110), so `builder.AddModulesFromAssembly(...).WithTags("x")` throws `InvalidOperationException("Module metadata must follow an AddModule call.")` at runtime (382-387). `AddModule(instance)` records the statically inferred type, so passing a variable typed as a base/interface tags the wrong type. A mutable-fluent footgun in a framework whose pitch is compile-time safety — and the fifth way to set tags/category.

**Fix:** return a typed registration handle: `builder.AddModule<TModule>()` -> `ModuleRegistration<TModule>` exposing `.WithTags(...)/.WithCategory(...)` (implicitly convertible back to the builder for chaining), deleting the last-registered state. Consider retiring registration-time tags entirely in favor of the module-owned surfaces.

Part of #3239 (round 2).

thomhurst:
Closes #3494

## Summary

- return `ModuleRegistration<TModule>` from typed `AddModule` overloads
- bind tags and categories to explicit registration handles, including concrete runtime instance types
- remove `LastRegisteredModuleType` and builder-level metadata extensions
- preserve existing fluent builder chains through facade forwarding and implicit builder conversion
- cover out-of-order handle configuration and runtime-type metadata

## Validation

- core solution build: passed, 0 warnings
- `PipelineBuilderRegistrationTests`: 19/19 passed

The unit-test project currently needs the one-line `RunnableModule` constructor fix from #3638 to compile; that fix was applied only for local validation and is not included here.

### DIFF
--- a/src/ModularPipelines.SourceGenerator/ModuleMetadataGenerator.cs
+++ b/src/ModularPipelines.SourceGenerator/ModuleMetadataGenerator.cs
@@ -26,6 +26,9 @@ public sealed class ModuleMetadataGenerator : IIncrementalGenerator
     internal const string PipelineBuilderExtensionsFullName =
         "ModularPipelines.Extensions.PipelineBuilderExtensions";
 
+    internal const string ModuleRegistrationNamespace = "ModularPipelines";
+    internal const string ModuleRegistrationTypeName = "ModuleRegistration";
+
     private static readonly DiagnosticDescriptor SkippedModuleRuntimeMetadata =
         GeneratorDiagnostics.SkippedModuleRuntimeMetadata;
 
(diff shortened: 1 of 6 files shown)


thomhurst:
Addressed both findings in `61a3019c15`:

- `ModuleMetadataGenerator` now recognizes both `PipelineBuilderExtensions.AddModule` and `ModuleRegistration<T>.AddModule`, preserving closed-generic metadata plus MPG0013/MPG0015 diagnostics throughout chains. Added a focused generator regression test.
- Removed `IDisposable` and the forwarded `Dispose()` method from registration handles.

Validation: core build 0 warnings; all 23 `ModuleMetadataGeneratorTests` passed; core unit project build 0 warnings; 19/19 registration tests passed (with the current-main #3638 fixes applied only for local compilation).

PR #3812, opened 2026-08-04
Require identity-aware run history lookup

### LINKED ISSUE #3745: v4: IRunHistoryStore — make the identity-aware lookup the single abstract read member
The default interface implementation of `GetLatestAsync(pipelineIdentity)` fetches the store's single latest report and returns null unless it *happens* to match the identity — so any custom store that implements only the two abstract members silently loses per-identity history the moment two pipelines share a backend, since the engine always calls the identity overload (`RunReportService.cs:81`). The docs acknowledge this trap ("its default implementation only validates the store's latest report"), which is a sign the API shape is wrong rather than the docs being incomplete.

**For v4:** make the identity-aware lookup the single abstract read member and delete the parameterless overload (or demote it to an extension method); this dovetails with the multi-run query reshape (see the "multi-run history queries" issue). Breaking for custom `IRunHistoryStore` implementations, which is exactly what a major version is for.

**Evidence:** `src/ModularPipelines/Engine/IRunHistoryStore.cs:23-31`, `docs/docs/how-to/run-reports.md:72-75`, `src/ModularPipelines/Extensions/PipelineBuilderExtensions.cs:309-317`

---

thomhurst:
## Summary
- make identity-aware `GetLatestAsync` the sole abstract history read API
- remove the ambiguous parameterless latest-report lookup from the built-in store
- simplify filesystem filtering and update custom-store documentation

## Validation
- `RunReportTests`: 42/42 passed
- focused filesystem history-store tests: 3/3 passed
- `ModularPipelines.slnx` Release build: 0 warnings/errors
- targeted whitespace verification and diff check

Fixes #3745

### DIFF
--- a/docs/docs/how-to/run-reports.md
+++ b/docs/docs/how-to/run-reports.md
@@ -69,7 +69,5 @@ register it on the builder:
 builder.AddRunHistoryStore<MyRunHistoryStore>();
 ```
 
-The store returns the latest report for comparison and saves the completed current report. Custom
-stores own their retention behavior. Override the identity-aware `GetLatestAsync` overload when the
-backend can search past a newer report belonging to another pipeline; its default implementation
-only validates the store's latest report.
+The store returns the latest report for the requested pipeline identity and saves the completed
+current report. Custom stores own their retention behavior.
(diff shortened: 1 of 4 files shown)


thomhurst:
Addressed Claude review in e51c16993f: the run-report guide now calls out the v4 `IRunHistoryStore` migration and names the sole identity-aware read member. @codex review

PR #3831, opened 2026-08-04
Publish run reports and history atomically

### LINKED ISSUE #3738: Run reports: history files written non-atomically; the service's own 30s timeout can truncate them
`FileSystemRunHistoryStore.SaveAsync` writes the history file with a single `File.WriteAllTextAsync` straight to its final path, and `RunReportService.SaveHistoryAsync` passes a 30-second timeout token into that write — so the framework itself can cancel mid-write, and a killed CI job or Ctrl+C can do the same. The result is a truncated JSON file that carries the newest timestamp in its name: it occupies a retention slot, fails deserialization with a logged warning on every subsequent run, and suppresses deltas until enough newer runs push it out.

`WriteReportAsync` has the same non-atomic pattern for the CI-consumed `artifacts/run-report.json`, so a CI artifact can be silently truncated with only a log warning.

**Fix:** write to a temp file in the same directory and `File.Move(temp, path, overwrite: true)` (atomic on the same volume on Windows and POSIX), deleting the temp file on failure; optionally delete files that repeatedly fail deserialization instead of warning forever.

**Evidence:** `src/ModularPipelines/Engine/FileSystemRunHistoryStore.cs:84-90`, `src/ModularPipelines/Engine/RunReportService.cs:152-159`, `src/ModularPipelines/Engine/RunReportService.cs:176-179`, `src/ModularPipelines/Engine/FileSystemRunHistoryStore.cs:62-65`

---

thomhurst:
Closes #3738.

## Summary
- write report JSON to a same-directory temporary file before atomic replacement
- use the same path for retained history files and CI report output
- remove unpublished temporary files after cancellation or failure without masking the primary error
- document atomic publication behavior

## Validation
- `AtomicFileWriter*`: 3 passed
- `RunReportTests`: 49 passed
- `ModularPipelines.slnx` Release build: 0 warnings, 0 errors
- scoped formatting passed


thomhurst:
Required pipeline failed on a shared timing-sensitive cancellation test unrelated to this PR. The workflow is now complete, so I triggered its single failed-job rerun.

thomhurst:
Required pipeline failed on two attempts due to unrelated shared EngineCancellationTokenTests timeouts. Refreshed with current main at

PR #3836, opened 2026-08-04
Exclude failed runs from duration deltas

### LINKED ISSUE #3741: Run reports: duration deltas compare against failed runs/modules, producing false regressions
`PipelineRunReportFactory.GetPreviousDuration` gates only on `DurationMeasured`, and a module that ran and *failed* after 2 seconds has a measured timeline (start+end both set in `CreateCurrentReportValues`), so the next successful 60-second run reports a "delta +58s" regression that never happened. Skipped/UsedHistory modules are correctly excluded via `DurationMeasured=false` (covered by `RunReportDoesNotCompareNonExecutedModuleDurations`), but failed/timed-out modules are not.

The same applies at pipeline level: `TotalDurationDelta` is computed against `previousReport.TotalDuration` unconditionally, and `LoadPreviousReportAsync` never filters by status, so a run after a fast-failing run shows a large spurious positive delta in the `Δ previous` column.

**Fix:** only compute module deltas when both current and previous status are `Successful` (or store previous status and let the printer annotate), and either skip pipeline-level deltas against failed runs or prefer the newest *successful* report for comparison.

**Evidence:** `src/ModularPipelines/Engine/PipelineRunReportFactory.cs:174-179`, `src/ModularPipelines/Engine/PipelineRunReportFactory.cs:79-82`, `src/ModularPipelines/Engine/PipelineRunReportFactory.cs:151-163`, `src/ModularPipelines/Engine/RunReportService.cs:69-90`

---

thomhurst:
## Summary
- calculate total-duration deltas only between successful pipeline reports
- calculate module deltas only between measured successful module executions
- retain successful-module comparisons even when the surrounding pipeline failed
- document successful-only comparison semantics

## Validation
- `RunReportTests`: 51/51 passed
- Failed/TimedOut current and previous regressions: passed
- post-main-sync core Release build: 0 warnings, 0 errors
- scoped format completed (expected F# unsupported notice only)

Closes #3741

### DIFF
--- a/docs/docs/how-to/run-reports.md
+++ b/docs/docs/how-to/run-reports.md
@@ -38,7 +38,9 @@ report is also exposed through `PipelineSummary.RunReport`.
 By default, the `IRunHistoryStore` saves reports under `.modularpipelines/run-history` on local and
 CI runs, even when JSON report writing is disabled. It retains the latest 20 reports and uses the
 newest compatible report to calculate module and total-duration deltas. When a previous duration
-exists, the final results table includes a `Δ previous` column.
+exists, the final results table includes a `Δ previous` column. Deltas compare only successful
+runs and successful module executions, so failed or timed-out durations do not create false
+regressions on a later run.
 
 Add the default history directory to `.gitignore` if you do not want to commit local run data:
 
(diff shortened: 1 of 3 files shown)


thomhurst:
@codex review

thomhurst:
Required pipeline failed on the first attempt due to unrelated shared-suite flake(s). Triggered the single failed-job rerun for workflow 30935410921.

PR #3846, opened 2026-08-04
Prune stale atomic report temp files

thomhurst:
Follow-up to #3831 review feedback.\n\n## Summary\n- prune owned atomic-write temp files older than 24 hours after successful history saves\n- retain recent temp files to avoid racing concurrent writers\n- leave unrelated temp files untouched\n- document the cleanup behavior\n\n## Validation\n- RunReportTests: 51 passed\n- focused post-main test: 1 passed\n- ModularPipelines.slnx Release build: 0 warnings, 0 errors


thomhurst:
@coderabbitai review

PR #3847, opened 2026-08-04
Obfuscate secrets in a single scan

### LINKED ISSUE #3753: Perf: obfuscation does one full pass over the input per secret pattern instead of a single SearchValues-driven scan
Once `ContainsAny` confirms at least one secret is present, `ObfuscateCaseSensitive` runs `StringBuilder.Replace(secret, mask)` for every registered pattern, and `ObfuscateCaseInsensitive` runs an `IndexOf` loop per pattern — each a full O(inputLength) pass even for patterns that never match. Because `SecretMaskingPatternGenerator` produces ~5 patterns per secret, a pipeline with 20 secrets performs ~100 full passes over the input; for a 1 MB captured command output (default `MaxCapturedOutputLength`) that is ~100 MB of char scanning per obfuscation call, multiplied by the repeated-obfuscation paths tracked separately.

**Fix:** after the `ContainsAny` gate, drive replacement with a single `IndexOfAny(SearchValues<string>)` loop (the `SearchValues` instance is already built and reports match positions), replacing longest-match-first at each hit — one pass total regardless of pattern count.

**Evidence:** `src/ModularPipelines/Engine/SecretObfuscator.cs:163-173,178-221`, `src/ModularPipelines/Engine/SecretMaskingPatternGenerator.cs:12-25`, `src/ModularPipelines/Options/CommandExecutionOptions.cs:70`

---

thomhurst:
## Summary
- replace per-secret full-input replacements with one SearchValues-driven forward scan
- preserve longest-match precedence for overlapping patterns
- avoid rescanning generated mask text

## Validation
- SecretObfuscatorCachingTests: 13 passed
- SecretMasking*: 36 passed
- ModularPipelines.slnx Release build: 0 warnings, 0 errors

Closes #3753

### DIFF
--- a/src/ModularPipelines/Engine/SecretObfuscator.cs
+++ b/src/ModularPipelines/Engine/SecretObfuscator.cs
@@ -16,14 +16,10 @@ namespace ModularPipelines.Engine;
 /// called concurrently from multiple threads without external synchronization.
 /// </para>
 /// <para>
-/// <b>Performance:</b> For optimal performance, secrets are sorted by length (longest first)
-/// to ensure longer secrets are masked before shorter ones that might be substrings.
-/// When case-insensitive matching is enabled, a single-pass algorithm using
-/// <see cref="StringComparison.OrdinalIgnoreCase"/> is used. This approach uses .NET's
-/// highly optimized string search which performs well for typical log message sizes.
-/// For extremely large log outputs with many secrets, consider reducing the number of
-/// registered secrets or using case-sensitive matching which uses the more efficient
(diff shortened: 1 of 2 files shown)


thomhurst:
Ubuntu pipeline failed only in the known retry timing race `When_Retry_With_Timeout_Then_Honour_Overall_Timeout` (expected one execution, observed two), unrelated to this PR's single-pass secret obfuscation. The root fix is pending in #3837. Triggered the single allowed failed-job rerun for workflow 30946543487.

PR #3864, opened 2026-08-05
Fix stale run-report test options

thomhurst:
## Summary
- remove stale `PrintLogo` and `PrintResults` initializers after those pipeline options were removed
- restore compilation for the core unit-test project and queued PR CI

## Validation
- `RunReportTests/BuilderRegistrationInvokesRunReportEnricher`: 1/1 passed
- targeted formatter passed

PR #3844, opened 2026-08-04
Respect non-throwing pipeline failure mode

thomhurst:
## Summary
- gate secondary exception propagation behind the same failure-throw policy as primary exceptions
- stop treating expected AlwaysRun dependency postponements as throwable secondary failures
- preserve the primary `ModuleFailedException` in fail-fast mode

## Validation
- `AlwaysRunTests`: 3 passed
- `PipelineExecutorTests`: 2 passed
- `ModularPipelines.slnx` Release build: 0 warnings, 0 errors

Closes #3786

PR #3868, opened 2026-08-05
Fix concurrent cache entry replacement

### LINKED ISSUE #3807: Caching: FileSystemModuleCache concurrent read/replace of the same entry throws IOException on Windows (readers lack FileShare.Delete)
`OpenReadAsync` opens entries with `FileShare.Read` only, while `WriteAsync` publishes with `File.Move(temporary, destination, overwrite: true)`. On Windows, replacing a file that another process has open without `FILE_SHARE_DELETE` fails, so two pipeline runs sharing one cache directory (or a run racing another run's save of the same fingerprint) surface `IOException` logged at Error level ("Error saving module result to module cache") even though nothing is actually wrong — the entry being written is byte-identical to the one being read.

**Fix:** open readers with `FileShare.Read | FileShare.Delete` so the atomic replace succeeds, and consider downgrading same-content publish races to Debug. The `File.Exists` check before opening is also a TOCTOU window that should be replaced by catching `FileNotFoundException`.

**Evidence:** `src/ModularPipelines/Caching/FileSystemModuleCache.cs:26-28,54`, `src/ModularPipelines/Engine/ModuleExecutionPipeline.cs:620-623`

---

thomhurst:
Closes #3807.

Allows cache readers to share deletion, removes the existence-check TOCTOU window, and uses Windows atomic replacement for entries that are currently open.

Validation:
- FileSystemModuleCacheTests: 2/2 passed on Windows
- lightweight core Release build: 0 warnings, 0 errors

### DIFF
--- a/src/ModularPipelines/Caching/FileSystemModuleCache.cs
+++ b/src/ModularPipelines/Caching/FileSystemModuleCache.cs
@@ -23,10 +23,25 @@ public sealed class FileSystemModuleCache : IModuleCacheStore
     {
         cancellationToken.ThrowIfCancellationRequested();
         var path = GetEntryPath(fingerprint);
-        Stream? stream = File.Exists(path)
-            ? new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, 64 * 1024, FileOptions.Asynchronous)
-            : null;
-        return Task.FromResult(stream);
+        try
+        {
+            Stream stream = new FileStream(
+                path,
+                FileMode.Open,
+                FileAccess.Read,
+                FileShare.Read | FileShare.Delete,
+                64 * 1024,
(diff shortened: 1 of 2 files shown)


thomhurst:
CI diagnosis: all three platform builds fail only at the shared RunReportTests.cs CS0117 regression. PR #3864 contains the focused fix. Fast-fail, analyzers, Native AOT, CodeQL, and review pass; no branch-specific failure found, so no duplicate patch or rerun here.

thomhurst:
The shared RunReportTests regression is now fixed on main by merged PR #3844. Refreshed this branch onto current main for a clean CI/review cycle.

thomhurst:
@codex review

thomhurst:
@claude review

PR #3929, opened 2026-08-10
Reject Bare values in grouped optional CLI options

### LINKED ISSUE #3783: CLI options: grouped optional-value options silently drop Bare entries mixed with explicit values
`AddGroupedOptionalValueOption` filters the collection with `Where(value => !value.IsBare)` twice: a collection `[CliOptionValue.Bare, "x"]` renders `--opt x` with the Bare occurrence silently discarded, and an all-Bare collection renders a single bare `--opt` regardless of how many Bare entries were supplied. Neither behavior is documented on `CliOptionAttribute.GroupValues`, and dropping user-supplied values without an error contradicts the strictness applied everywhere else in the builder (empty values throw).

Since a bare occurrence inside a grouped value list has no meaningful rendering, the honest behavior is to throw an `InvalidOperationException` naming the type and property when a grouped optional collection contains Bare alongside values (or at all), rather than silently rewriting the command.

**Evidence:** `src/ModularPipelines/Helpers/Internal/CommandArgumentBuilder.cs:222-249`

---

thomhurst:
## Summary
- reject CliOptionValue.Bare in grouped optional-value collections instead of silently dropping it
- name the offending options type and property in the exception
- cover Bare-only and mixed Bare/explicit collections

## Test plan
- [x] CliAttributeTests (81/81)
- [x] ModularPipelines.slnx Release build (0 warnings, 0 errors)
- [x] whitespace verification

Closes #3783

### DIFF
--- a/src/ModularPipelines/Helpers/Internal/CommandArgumentBuilder.cs
+++ b/src/ModularPipelines/Helpers/Internal/CommandArgumentBuilder.cs
@@ -412,10 +412,15 @@ internal sealed class CommandArgumentBuilder : ICommandArgumentBuilder
                 $"Grouped option '{GetEffectiveName(optionPart.Attribute)}' must use a space separator.");
         }
 
+        if (optionValues.Any(static value => value.IsBare))
+        {
+            throw new InvalidOperationException(
+                $"Grouped optional-value CLI option property '{optionsType.FullName}.{optionPart.PropertyName}' "
+                + $"cannot contain {nameof(CliOptionValue)}.{nameof(CliOptionValue.Bare)}.");
+        }
+
         args.Add(GetEffectiveName(optionPart.Attribute));
-        args.AddRange(optionValues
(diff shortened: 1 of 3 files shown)


thomhurst:
CI audit: the Ubuntu runner lost communication with GitHub; no test or build failure was reported. Re-running the failed job once; no code changes.

thomhurst:
CI fix pushed at exact head aea85b1f336c4073ba6123b866b2a367c0b7738c. The rerun exposed AnsiConsole output from concurrent ExecutionHintTests inside SpectreResultsPrinterTests; the class now uses global TUnit NotInParallel isolation. Focused class: 10/10 green after rebase. Full unit-test project reached the mandated 300-second local guard without reporting failures; limit was not raised.

@codex review exact head aea85b1f336c4073ba6123b866b2a367c0b7738c

@claude review exact head aea85b1f336c4073ba6123b866b2a367c0b7738c and submit approval if clean.

PR #4009, opened 2026-08-23
Prevent casing-variant generated file duplicates

### LINKED ISSUE #3991: Generate CLI Options: regeneration emits casing-variant duplicate .Generated.cs files causing NETSDK1022 (gh, kind, kubectl)
In every scheduled **Generate CLI Options** run since 2026-08-10 (latest: [run 32613138952](https://github.com/thomhurst/ModularPipelines/actions/runs/32613138952)), the post-generation build of three tool packages fails with:

```
error NETSDK1022: Duplicate 'Compile' items were included.
```

Duplicates are casing variants of the same generated type — the regeneration writes a newly-cased filename while the previously-cased file still exists on disk (case-sensitive Linux runner keeps both):

- **gh** (`ModularPipelines.GitHub`): `GhAgentTask` / `GhGpgKey` / `GhSshKey` + interfaces `IGhGpgkey` / `IGhSshkey` (note interface vs class casing mismatch in the same run)
- **kind**: `KindExportKubeConfigOptions` / `KindGetKubeconfigOptions`
- **kubectl** (`ModularPipelines.Kubernetes`): `KubernetesClusterinfo` / `IKubernetesClusterinfo`

This looks like fallout from the compound-word PascalCase rename work (#3320): the casing normalizer produces a different name than what is on disk (and is itself inconsistent — `GhGpgKey.Generated.cs` alongside `IGhGpgkey.Generated.cs`), and the regeneration step does not delete superseded files before writing.

**Fix:**
1. Make the regeneration step clean the target Generated directory (or delete files it is about to re-emit under a different casing) before writing.
2. Fix the casing inconsistency so class and interface for the same service agree (`GpgKey` vs `Gpgkey`, `ClusterInfo` vs `Clusterinfo`, `KubeConfig` vs `Kubeconfig`).

Separate observation from the kubectl job worth fixing while in there: the scraper probes `kubectl --version`, which current kubectl rejects (`error: unknown flag: --version`); it should use `kubectl version`.

Part of the Generate CLI Options workflow outage tracked in the umbrella issue.

thomhurst:
Closes #3991.

## Summary
- compare generated keep-paths with the output filesystem's actual case sensitivity, so Linux prunes stale casing variants instead of preserving both
- normalize `AgentTask`, `ClusterInfo`, `GpgKey`, `SshKey`, and existing `KubeConfig` compound identifiers consistently across options, services, and interfaces
- probe kubectl with `version --client`, avoiding both the rejected `--version` flag and live-cluster dependency
- add orchestration, naming, service-generation, and kubectl probe regressions

## Validation
- OptionsGenerator tests: 823 passed
- OptionsGenerator Release build: 0 warnings, 0 errors
- scoped whitespace format verification passed
- local kubectl 1.36.1: `version` exits 1 without a cluster; `version --client` exits 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated API compatibility when options are removed, renamed, or changed.
  * Preserved compatibility aliases, facades, constructors, and documentation references.
  * Improved collision handling and cleanup of stale files across case-sensitive and case-insensitive filesystems.
  * Corrected command naming, enum handling, optional options parameters, and secret-option detection.
* **Tests**
  * Added comprehensive coverage for compatibility preservation, naming conversions, aliases, collisions, and output cleanup.


thomhurst:
Addressed the availability-probe review in 594b3a0f7f. The executor contract now accepts tool-specific probe arguments; the process executor runs that probe, while the resilient decorator delegates it directly to the inner executor so an open shared circuit cannot reject availability checks. Helm and kubectl now use the shared base implementation, removing both duplicate overrides. Added scraper coverage plus a regression proving the probe bypasses shielded execution. Validation: OptionsGenerator tests 824/824; Release build 0 warnings, 0 errors.

thomhurst:
Addressed the remaining filesystem-probe note in d394d1249e. GenerateAsync now probes the output directory once and threads that comparer through every tool; GenerateFromDefinitionAsync reuses the comparer it already computed. This avoids hidden persistent cache state. Validation: OptionsGenerator tests 824/824; Release build 0 warnings, 0 errors.

thomhurst:
Rebased onto current main (13fd094b55). Focused validation: orchestrator 16/16; hardening 52/52; GeneratorUtils 132/132; Helm 2/2; Kubectl 1/1; resilient executor 1/1. OptionsGenerator Release build clean.

thomhurst:
Rebased onto current main after #4023/#3985 and force-pushed with an exact lease. Validation: orchestrator, generator hardening/utilities, Helm, Kubectl, and resilient executor suites passed (204/204); OptionsGenerator Release build succeeded with 0 warnings/errors; diff checks clean. Verified remote head: 066cf5b199a84438a09e67ca19c7f790c650f8bd.

thomhurst:
Addressed the nullable alias-constructor finding in c0c40ae2bf. Alias wrappers now preserve the source type's nullable suffix for both enum and non-enum parameters; focused tests and the full OptionsGenerator suite pass (916/916), and the Release build is clean.

I verified the optional-facade finding against the existing end-to-end regression. `GeneratedApiCompatibilityPreserver` restores the baseline parameterless compatibility constructor when a previously optional facade gains a required member, so `options ?? new ToolAddOptions()` remains compilable. Changing `RequiresOptionsParameter` as suggested breaks that compatibility regression and causes the facade validation to reject the preserved method; no change retained for that finding.

Issue #3996, opened 2026-08-23
Generate CLI Options workflow red for all ~24 tools since 2026-08-10 — generated options frozen ahead of v4


thomhurst:
The scheduled **Generate CLI Options** workflow has failed on every run since 2026-08-10 (latest: [run 32613138952](https://github.com/thomhurst/ModularPipelines/actions/runs/32613138952)), with all ~24 tool jobs red: argocd, aws, az, brew, cosign, dotnet, flyway, gcloud, gh, git, go, kind, kubectl, liquibase, newman, packer, pip, pnpm, podman, pulumi, snyk, terraform, vault, winget.

The failures are the new strict generation validation (from the v4 audit work) catching real parser gaps and drift — the guards are doing their job, but the net effect is that **no tool options have regenerated for two weeks**. Since v4 renamed the entire generated surface, we want at least one fully green regeneration pass before tagging v4.0.0 so packages ship from fresh scrapes.

## Failure classes (one child issue each)

- [x] #3986 — aws help scraping produces zero commands (whole package cannot regenerate)
- [x] #3987 — positional operands declared in usage but no `CliPositionalArgument` generated (winget, dotnet, go, liquibase, pip, pnpm)
- [x] #3988 — options documented as repeatable parsed as scalar (terraform, pip, packer, snyk)
- [x] #3989 — az explicit true/false options parsed as presence-only flags (~12 options)
- [x] #3990 — command groups yield zero child commands (cosign, pulumi env run, snyk test)
- [x] #3991 — regeneration emits casing-variant duplicate `.Generated.cs` files → NETSDK1022 (gh, kind, kubectl); likely #3320 rename fallout
- [x] #3992 — documentation example catalogs reference missing commands (newman, terraform)
- [x] #3993 — command-coverage sentinel validation failing for ~9 tools (triage after the above: some downstream, flyway looks like genuine drift)
- [x] #3994 — gcloud runner shutdown exit 143; podman job fails despite generator reporting zero errors

## Done when

A scheduled (or manually dispatched) Generate CLI Options run completes green for every tool.

Suggested order: #3987/#3988/#3989/#3990 (shared parser gaps, unblock most tools) → #3986 (aws adapter) → #3991 (regen cleanup + casing) → #3992 → re-run workflow → #3993/#3994 triage of whatever remains.

thomhurst:
[...]

PR #4188, opened 2026-08-26
Preserve CLI option generation safety

thomhurst:
## Summary
- add explicit Cosign type overrides for identity tokens and OIDC client secret files
- preserve masking after type enhancement recognizes path-capable values
- scope Homebrew option parsing to the selected usage block so wrapper help cannot leak unrelated options
- retain full-help fallback when usage matching fails and stop matched slices at the next usage block
- share usage-location logic between option and description parsing

## Validation
- exact-head CI OptionsGenerator suite: 1,186 passed; generated global options verified
- local OptionTypeEnhancerTests Release: 9 passed
- local BrewCliScraperTests Release: 15 passed
- OptionsGenerator Release build: 0 warnings, 0 errors
- scoped whitespace format and diff checks passed
- Homebrew 6.0.19 live help reproduced the rubocop/install-bundler-gems dual-usage boundary

Continues #3996.


thomhurst:
Addressed Claude's actionable findings in d39580b19b: unmatched usage now falls back to raw help, matched slices stop at the next Usage block, description/options share one matching locator, and leading/trailing plus fallback regressions pass in the 15-test Brew suite.

PR #4205, opened 2026-08-26
Fix circular obsolete guidance in CLI facades

### LINKED ISSUE #4204: Fix circular obsolete guidance in restored CLI facades
## Problem

Automated Homebrew regeneration in #4202 exposed circular deprecation guidance for commands that changed from nested CLI tokens to required operands. Generated PinInstalled_formulaAsync points to PinInstalledFormulaAsync, but both methods are compatibility-only and obsolete. The same occurs for unpin.

The durable fix belongs in GeneratedApiCompatibilityPreserver, not generated Homebrew files.

## Acceptance criteria

- Detect when a removed nested command maps to a live command whose required operand represents the removed suffix.
- Point restored facade aliases to the live non-obsolete facade, such as PinAsync or UnpinAsync.
- Never emit Use-X guidance that targets a compatibility-only method.
- Add focused generator regression tests.
- Regenerate affected integrations only through the normal automated workflow.

thomhurst:
## Summary

- detect restored nested commands that became required operands
- point legacy facade aliases to the live root facade
- fall back to the compatibility-only warning instead of targeting another obsolete method
- add regression coverage for the Homebrew pin shape

Generated integration files are intentionally unchanged; the normal automated workflow will regenerate them after this generator fix lands.

## Validation

- OptionsGenerator tests: 1,229 passed with Cobertura coverage
- OptionsGenerator Release build: passed, 0 warnings
- changed-file whitespace verification: passed
- whole-solution format verification: blocked by pre-existing whitespace/analyzer debt outside this change

Closes #4204

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility guidance for removed commands by identifying the most relevant replacement command.
  - Preserved accurate guidance for nested commands and required positional arguments.
  - Added safeguards to fall back to the compatibility message when no unique replacement can be determined.
  - Corrected messaging to recommend `PinAsync` where appropriate.

### DIFF
--- a/tools/ModularPipelines.OptionsGenerator/src/ModularPipelines.OptionsGenerator.Tests/Generators/GeneratorHardeningTests.cs
+++ b/tools/ModularPipelines.OptionsGenerator/src/ModularPipelines.OptionsGenerator.Tests/Generators/GeneratorHardeningTests.cs
@@ -5042,7 +5042,20 @@ public class GeneratorHardeningTests
                 "namespace ModularPipelines.Tool.Services; public class Tool { "
                 + "public Task PinInstalled_formulaAsync(ToolPinInstalled_formulaOptions? options = null) "
                 + "=> Task.CompletedTask; }");
-            var current = Command("ToolPinOptions", "ToolOptions", ["pin"]);
+            var current = Command("ToolPinOptions", "ToolOptions", ["pin"]) with
+            {
+                PositionalArguments =
+                [
+                    new CliPositionalArgument
+                    {
+                        PropertyName = "InstalledFormula",
+                        CSharpType = "IEnumerable<string>",
+                        IsRequired = true,
+                        PositionIndex = 0,
+                        Phase = CommandLinePhase.Passthrough,
+                    },
+                ],
+            };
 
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed Claude efficiency suggestion in 8961a839e5: live root-facade candidates are now computed once in RestoreRemovedCommands and reused for each restored command. Full OptionsGenerator suite (1,229 tests), Release build, and changed-file whitespace verification pass.

thomhurst:
Follow-up disposition:

1. No ambiguous-match test added: for a fixed removed command path there is only one prefix at each length. Two same-length candidates must have the same path (or casing variant), and `GetNonCollidingRootCommands` rejects their duplicate normalized root method before this helper runs. The defensive `Take(2)` remains fail-closed.
2. No shared uniqueness helper: the cited sites intentionally have different contracts (`null` for no safe replacement versus exception for ambiguous baseline identity). Keeping those semantics local is clearer.
3. No extra caching: the suffix identifier depends on each candidate prefix length and is computed only within the required live-command scan. Current complexity is acceptable after hoisting the invariant root-facade set.

No unresolved actionable concern remains.

PR #4431, opened 2026-08-31
Assert legacy Git context accessor removal

thomhurst:
## Summary

- replace the stale assertion that `GitExtensions.Git()` is hidden/obsolete
- assert the legacy context accessor is absent, matching the v4 API decision from #4404
- unblock full-pipeline Git tests on current `main`

## Validation

- `ModularPipelines.Git.UnitTests`: 52 passed (Release, net10.0)

Part of #4331.

PR #4460, opened 2026-08-31
Classify secret values without masking identifiers

### LINKED ISSUE #4329: OptionsGenerator: classify secret values without masking identifiers
## Problem

PR #4292 generates `[SecretValue]` on `GhSecretSetOptions.SecretName`, which is only an identifier, while leaving the actual `--body` secret value unmarked. Command logging can therefore reveal secret material and redact the harmless name.

Root cause: `GeneratorUtils.IsSecretOption` relies primarily on property-name keywords. It treats `SecretName` as secret because it contains `Secret`, but misses generic value-bearing names such as `Body` even when their CLI description says "The value for the secret". The same false-positive pattern can affect `*TokenName`, `*KeyId`, and similar identifiers.

## Acceptance criteria

- Secret detection distinguishes identifiers (`*Name`, `*Id`) from secret-bearing values.
- Description/context signals can mark generic value properties such as `Body` or `Value` when they carry secret material.
- `GhSecretSetOptions.Body` receives `[SecretValue]`; `SecretName` does not.
- `GhSecretDeleteOptions.SecretName` is not marked secret.
- Add generic generator regression coverage plus GitHub CLI scraper coverage.
- Regenerate GitHub options; do not hand-edit generated files.

Discovered while validating #4292.

thomhurst:
## Summary
- exclude identifier-shaped properties such as `SecretName`, `TokenName`, and `*Id` from secret masking
- classify generic values only when descriptions directly identify secret material
- regenerate GitHub CLI v2.98.0 so `secret set --body` is masked while secret names remain visible
- retain current generated command surface, including `gh stack`, and merged v4 generator drift

## Validation
- `ModularPipelines.OptionsGenerator.slnx` Release build
- `ModularPipelines.GitHub.slnx` Release build (19 pre-existing warnings, 0 errors)
- 146 `GeneratorUtilsTests` passed
- 19 `ZeroOutputScraperTests` passed
- `git diff --check` passed

Closes #4329

### DIFF
--- a/docs/docs/mp-packages/cli/gh.md
+++ b/docs/docs/mp-packages/cli/gh.md
@@ -254,6 +254,7 @@ public class RunCommandModule : Module<CommandResult>
 | `gh ssh-key add` | `GhSshKeyAddOptions` |
 | `gh ssh-key delete` | `GhSshKeyDeleteOptions` |
 | `gh ssh-key list` | `GhSshKeyListOptions` |
+| `gh stack` | `GhStackOptions` |
 | `gh status` | `GhStatusOptions` |
 | `gh variable` | `GhVariableOptions` |
 | `gh variable delete` | `GhVariableDeleteOptions` |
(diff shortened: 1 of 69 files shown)


thomhurst:
Rebased onto current main to pick up #4431's fix for the unrelated LegacyContextAccessorIsHiddenAndObsolete CI failure. Exact head: a2efd2bf5ca609fe36e0737811edc22ce44b2b5d

PR #4490, opened 2026-09-01
Stabilize archive extraction cancellation test

### LINKED ISSUE #4479: Stabilize archive extraction cancellation test
## Problem

The Ubuntu pipeline can fail in `ArtifactContextApiTests.Directory_Archive_Extraction_Observes_Cancellation` because the 64 MiB no-compression extraction completes before `CancellationTokenSource(TimeSpan.FromMilliseconds(1))` fires. The assertion then expects `OperationCanceledException`, but no exception is thrown.

Observed on PR #4421 at commit `83f529847fef46f0e7c614c727db07d98dc6d3e7`, which changes only Terraform documentation.

Failed rerun: https://github.com/thomhurst/ModularPipelines/actions/runs/33448749830/job/99822014091

## Acceptance criteria

- Remove wall-clock/runner-speed dependence from `Directory_Archive_Extraction_Observes_Cancellation`.
- Preserve meaningful coverage that directory archive extraction observes cancellation, including the intended cancellation checkpoint.
- Keep temporary archive/destination cleanup reliable on assertion failure.
- Add repetition coverage or otherwise demonstrate deterministic behavior on Linux.

thomhurst:
Closes #4479

Replaces the runner-speed-dependent timer cancellation with a deterministic stream checkpoint, repeats the scenario ten times, and guarantees cleanup after the extraction task settles.

Validation: ArtifactContextApiTests 19/19 passed; formatting verification passed.

### DIFF
--- a/test/ModularPipelines.UnitTests/Api/ArtifactContextApiTests.cs
+++ b/test/ModularPipelines.UnitTests/Api/ArtifactContextApiTests.cs
@@ -331,36 +331,53 @@ public class ArtifactContextApiTests
     [Test]
     public async Task Directory_Archive_Extraction_Observes_Cancellation()
     {
-        var archivePath = Path.Combine(Path.GetTempPath(), $"{Guid.NewGuid():N}.zip");
-        var destinationDirectory = Directory.CreateTempSubdirectory("artifact-extraction-");
+        await using var archiveStream = new MemoryStream();
+        using (var archive = new ZipArchive(archiveStream, ZipArchiveMode.Create, leaveOpen: true))
+        await using (var entryStream = archive.CreateEntry(
+                         "payload.bin",
+                         CompressionLevel.NoCompression).Open())
+        {
+            await entryStream.WriteAsync(new byte[1024 * 1024]);
+        }
+
+        var archiveBytes = archiveStream.ToArray();
+        for (var iteration = 0; iteration < 10; iteration++)
+        {
+            await VerifyArchiveExtractionCancellationAsync(archiveBytes, iteration);
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude review

thomhurst:
@codex review

thomhurst:
Ubuntu failed in the unrelated Timeout_Fault_During_Grace_Period_Counts_As_Response test (from current main), not the archive cancellation changes. PR #4476 contains the timeout race fix and is now under fresh CI. I am leaving this branch unchanged until that fix lands; then this PR can refresh from main or receive a confirmed-flake rerun.

thomhurst:
Fixed in bf941d6a5d. The cancellation assertion now awaits extractionTask through a five-second WaitAsync bound, so a cancellation-forwarding regression reaches the existing finally block, releases the blocked read, settles the task, and cleans the temporary directory. The exact test passes (1/1, with ten internal repetitions); scoped whitespace verification passes.

thomhurst:
Addressed the exact-head Claude review in 1168448fd5. The archive helper is now explicitly named SwitchableBlockingReadStream to distinguish its seekable delegating/toggle behavior from the synthetic HTTP response helper. Both synchronous Read overloads participate in the gate; the cross-thread flag is volatile; extraction runs off-thread so an early synchronous ZipArchive read cannot block test setup; the blocking cancellation token interrupts sync reads; and the finally cancellation is documented as defensive cleanup for early setup/assertion failures. Static diff checks pass. Final executable validation was deferred to CI because the guarded local test process exceeded the mandated 2 GB cap (exit 137); limits were not raised or retried.

PR #4506, opened 2026-09-02
Keep distributed retries node-local

### LINKED ISSUE #4382: Distributed: retry configuration not portable — assignments hardcode RetryCount 0
## Problem

Module retry configuration does not survive the wire. `DistributedWorkPublisher.CreateAssignment` hardcodes `RetryCount: 0` when building `ModuleAssignmentConfig` (`src/ModularPipelines/Distributed/Master/DistributedWorkPublisher.cs:64,73`), so a module that declares retries (via `RetryConfiguration`, `ResilienceShieldFactory`, or `PipelineOptions.DefaultRetryCount`) runs with zero retries when assigned to a worker.

The same limitation applies to the richer resilience configuration: `ResilienceShieldFactory` is a delegate and inherently non-portable, but the simple declarative `RetryCount` case can and should be transmitted.

## Proposed change

- Populate `RetryCount` (and the effective timeout, which is already sent) from the module's resolved configuration when creating the assignment.
- Document that delegate-based resilience shields are evaluated on the executing node from the module's own `Configure` — verify that the worker-side execution pipeline actually applies them (the worker rebuilds configuration from the module type, so declarative config may already work; the hardcoded assignment value is then misleading and should be removed rather than duplicated).
- Add a distributed test that a retrying module actually retries on a worker.

Part of #4372.

thomhurst:
## Summary
- remove unused `RetryCount` from the unshipped distributed assignment configuration
- document retry policy resolution on the executing node
- verify workers apply module `WithRetry`, custom `Shield` factories, and `PipelineOptions.DefaultRetryCount`
- update coordinator test fixtures and the unshipped API baseline

The wire field was never consumed. Keeping retry policy node-local preserves complete behavior, including predicates and delegate-based shields, instead of transmitting a misleading integer subset.

## Validation
- `ModularPipelines.Distributed.UnitTests` Release: 122 passed with coverage
- focused worker retry tests: 3 passed with coverage
- `ModularPipelines.slnx` Release: 0 warnings, 0 errors
- Distributed Redis solution Release: 0 warnings, 0 errors
- Distributed SignalR solution Release: 0 errors; one existing warning in untouched `MasterServerHost.cs:108`
- scoped whitespace verification: passed

Closes #4382

## Summary by CodeRabbit

* **New Features**
  * Distributed workers now resolve retry policies locally from module settings or the pipeline’s default retry configuration.
  * Retry behavior supports module-defined policies, resilience shield factories, and default retry settings.

* **Breaking Changes**
  * Removed the `RetryCount` property and constructor parameter from distributed module assignment configuration.

* **Tests**
  * Added coverage validating retry behavior across module, shield-factory, and default configurations.

### DIFF
--- a/src/ModularPipelines/Distributed/Master/DistributedWorkPublisher.cs
+++ b/src/ModularPipelines/Distributed/Master/DistributedWorkPublisher.cs
@@ -62,7 +62,6 @@ internal class DistributedWorkPublisher(
 
         var dependencyResults = GatherDependencyResults(module);
 
-        // Distributed workers do not yet consume portable retry configuration.
         return new ModuleAssignment(
             ModuleTypeName: moduleType.FullName!,
             ResultTypeName: resultTypeName,
@@ -70,7 +69,6 @@ internal class DistributedWorkPublisher(
             AssignedAt: DateTimeOffset.UtcNow,
             Configuration: new ModuleAssignmentConfiguration(
                 TimeoutSeconds: config.Timeout is not null ? (int?) config.Timeout.Value.TotalSeconds : null,
-                RetryCount: 0,
                 AlwaysRun: config.AlwaysRun
             ),
             DependencyResults: dependencyResults)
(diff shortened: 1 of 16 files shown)


thomhurst:
Rebased onto current main at 7bf4d3efff, incorporating #4476's fix for the unrelated Ubuntu failure Timeout_Fault_During_Grace_Period_Counts_As_Response. Distributed tests pass 122/122 after rebase.

PR #4534, opened 2026-09-03
Isolate collectible metadata leak probes

### LINKED ISSUE #4471: Stabilize collectible metadata unload tests on Linux
## Problem

`GeneratedRuntimeMetadataTests.GeneratedMetadata_DoesNotRootCollectibleAssemblies` intermittently leaves the collectible dynamic assembly/type alive on Ubuntu CI after ten back-to-back `GC.Collect` / `WaitForPendingFinalizers` cycles.

Observed in #4464's `.NET` run: https://github.com/thomhurst/ModularPipelines/actions/runs/33451719954/job/99689360978

The failing assertion was `typeReference.IsAlive == true`. This is unrelated to #4464's generated whitespace diff. The test has no yield or delay between collection attempts, unlike other collectible/reference tests in this repository, so it cannot currently distinguish a real metadata registry root from delayed runtime/JIT cleanup.

## Acceptance criteria

- Determine whether generated metadata still holds a strong reference or the failure is collection timing.
- Make both collectible metadata tests deterministic on Linux without weakening their leak assertions.
- Add stress/repetition coverage that fails for a real strong root and passes for delayed unload.
- Keep CI from requiring blind reruns for this failure.

Discovered while maintaining #4464.

thomhurst:
## Summary
- give collectible metadata leak probes process-wide test isolation instead of a key shared only by the three probes
- keep the dedicated registration thread cancellation-aware so a timeout cannot strand the test method in `Thread.Join`
- remove unrelated global metadata-table enumeration from the external leak probe; lookup behavior remains covered by dedicated registration tests

## Validation
- `ModularPipelines.Tests.slnf` Release build: passed (0 errors; 181 existing warnings)
- `GeneratedRuntimeMetadataTests`: 33/33 passed in 5 consecutive processes (165 total)
- scoped whitespace verification: passed
- full core test project: local validation stopped by the mandated 2 GB process-tree guard at 2456 MB; limit was not raised or retried

Closes #4471

### DIFF
--- a/test/ModularPipelines.UnitTests/Attributes/GeneratedRuntimeMetadataTests.cs
+++ b/test/ModularPipelines.UnitTests/Attributes/GeneratedRuntimeMetadataTests.cs
@@ -675,7 +675,7 @@ public class GeneratedRuntimeMetadataTests
     }
 
     [Test]
-    [TUnit.Core.NotInParallel("CollectibleMetadata")]
+    [TUnit.Core.NotInParallel]
     [Timeout(30_000)]
     public async Task GeneratedMetadata_DoesNotRootCollectibleAssemblies(
         CancellationToken cancellationToken)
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude review

Issue #4331, opened 2026-08-30
v4: remove generated API compatibility preservation; tool output is source of truth


thomhurst:
## Problem

Generated integration APIs are being shaped by earlier generated C# instead of the current tool. `GeneratedApiCompatibilityPreserver` reads previous output and restores removed commands, properties, constructors, facade methods, aliases, and enums. `DependencyRegistrationGenerator` also emits old `context.Tool()` accessors.

Current evidence:

- 4,705 `[Obsolete]` declarations in `*.Generated.cs`.
- 1,101 generated members say they have no effect or exist only for compatibility.
- 58 obsolete `context.Tool()` accessors remain across generated and handwritten integrations.
- The current obsolete count is effectively unchanged from v3.2.1285.

This conflicts with the v4 policy: generated options follow the installed CLI/tool. Generated integration packages do not promise API compatibility when their upstream tool changes.

## Decision

Current scraper/tool output is the sole source of truth. Breaking generated API drift is accepted. Do not retain a command, option, property name, constructor, facade, enum member, or accessor only because an earlier ModularPipelines version generated it.

Normalization required to model the current tool accurately may remain. Historical ModularPipelines API preservation may not.

## Scope

- Remove `GeneratedApiCompatibilityPreserver` and its call sites from `CodeGeneratorOrchestrator`.
- Remove historical compatibility models and emitter branches used only to restore earlier generated APIs (`CompatibilityProperties`, compatibility-only commands/methods/enums/constructors, and equivalent paths).
- Stop reading existing generated C# or API baselines to add members to a new generation result.
- Stop emitting traditional `context.Tool()` compatibility accessors; retain registration and current `context.Tools.*` discovery.
- Remove handwritten compatibility accessors following the same old pattern.
- Regenerate every tool integration and refresh generated API baselines from the resulting source.
- Update generator tests: assert current tool models exactly, not backward-compatible API shapes.

## Acceptance criteria

- Generated output contains only members represented by the current scraped/supplemental tool model.
- Removed upstream commands and options disappear rather than becoming obsolete no-op members.
- Renamed/collision-resolved members do not retain aliases solely for ModularPipelines compatibility.
- No `context.Tool()` compatibility accessors remain; pre-C# 14 callers use `context.Tools.Get<T>()`.
- Generation never consults previous generated output or public API baselines to restore symbols.
- Obsolete annotations remain only when they describe something still present in the current upstream tool model, not prior ModularPipelines output.
[...]

PR #4537, opened 2026-09-03
Refresh eksctl CLI options from current output

thomhurst:
Part of #4331.

## Summary
- regenerate eksctl options from current eksctl v0.230.0 output
- remove compatibility members no longer present in the CLI
- update the enum rendering test for the generated `ControllerManager` member

## Validation
- eksctl solution build: 0 warnings, 0 errors
- eksctl tests: 7/7 passed
- `git diff --check`
- scoped formatter reports only the repository's pre-existing baseline findings


thomhurst:
@claude Please review this current-output eksctl regeneration.

PR #4544, opened 2026-09-03
Avoid generated validation member collisions

thomhurst:
Part of #4331.

Refreshes the Kubernetes integration from current kubectl v1.37.0 output and fixes a generator member collision exposed by `kubectl apply --validate`.

Changes:
- emits required-alternative validation through explicit `IValidatableObject.Validate` implementation
- adds regression coverage for a CLI option named `Validate`
- regenerates `KubernetesApplyOptions` and its service signatures from kubectl v1.37.0
- updates the rendering fixture to satisfy the current filename-or-kustomize requirement

Validation:
- generator regression test: passed
- Kubernetes solution build: 0 warnings, 0 errors
- Kubernetes unit tests: 11 passed
- `git diff --check` clean


thomhurst:
@claude Please review this kubectl synchronization and generator collision fix.

PR #4597, opened 2026-09-03
Stabilize distributed routing timing tests

### LINKED ISSUE #4593: Stabilize distributed capability-routing timing tests
Multiple otherwise unrelated PRs fail the distributed suite under hosted-runner contention:

- `Executor_Fails_Impossible_Capability_After_Registration_Grace` asserts completion under two seconds even though its `[Timeout]` already provides the hang bound.
- `Capability_Routing_Grace_Starts_When_Assignment_Is_Ready` relies on the test continuation publishing a worker within a 300 ms production grace window.

Observed in PR CI runs #4367 and #4579. Make both tests deterministic without weakening their behavioral assertions.

thomhurst:
## Summary
- remove the redundant two-second wall-clock ceiling from the impossible-capability test; its TUnit timeout still bounds hangs
- make worker registration advance on coordinator queries instead of relying on the test continuation winning a 300 ms scheduling race
- retain assertions that the grace period is observed and queried more than once

## Validation
- `DistributedModuleExecutorTests`: 54/54 passed
- repeated focused class 5 additional times: 270/270 passed
- scoped format verification and `git diff --check`: clean

Fixes #4593

### DIFF
--- a/test/ModularPipelines.Distributed.UnitTests/Master/DistributedModuleExecutorTests.cs
+++ b/test/ModularPipelines.Distributed.UnitTests/Master/DistributedModuleExecutorTests.cs
@@ -2565,7 +2565,6 @@ public class DistributedModuleExecutorTests
         using (Assert.Multiple())
         {
             await Assert.That(sw.Elapsed).IsGreaterThanOrEqualTo(TimeSpan.FromMilliseconds(50));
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude Please review exact head bb7e2f0887220b499d98f2cd623acb28d0db0c7b. This test-only change removes two hosted-runner timing races observed in #4367 and #4579.

Issue #4655, opened 2026-09-06
Use the shared continuation rule in CliScraperBase.HelpDeclaresRepeatableOption


thomhurst:
Part of #4646.

`CliScraperBase.HelpDeclaresRepeatableOption` walks continuation lines with `!OptionLinePattern().IsMatch(...)`, a third continuation rule next to `IsContinuationLine`/`AccumulateWrappedDescription`. Rebase that lookahead on the shared helper so a wrapped line that looks like an option row is still part of the option's documentation when checking for repeatable-value prose.

### Acceptance criteria
- `HelpDeclaresRepeatableOption` uses `AccumulateWrappedDescription` (or `IsContinuationLine`) instead of its own stop rule.
- A test covers a repeatable-value phrase that only appears on a wrapped line beginning with `--flag`.
- `ModularPipelines.OptionsGenerator.Tests` passes.

Issue #4646, opened 2026-09-05
Port the remaining CLI scrapers onto the shared wrapped-description helper


thomhurst:
Follow-up to #4634 / #4645.

#4645 hoisted the column-aware wrapped-description mechanism onto `CliScraperBase` (`GetIndentation`, `GetColumn`, `IsContinuationLine`, `AccumulateWrappedDescription`) and collapsed the Cargo, Git, Cobra and Hadolint variants onto it. Fourteen scrapers still carry a private `AccumulateMultiLineDescription` with a hard-coded absolute indent threshold (or a plain regex `IsMatch` stop), so they retain the defect class the issue targeted: a wrapped line whose first token is `--foo` is treated as a new option and dropped from the description.

Remaining scrapers under `tools/ModularPipelines.OptionsGenerator/src/ModularPipelines.OptionsGenerator/Scrapers/Cli/`:

- dash-prefix + absolute threshold: `AzCliScraper` (20), `DotNetCliScraper` (10), `GoCliScraper` (8), `MavenCliScraper` (30), `PipCliScraper` (20), `PnpmCliScraper` (20), `TerraformCliScraper` (10), `WinGetCliScraper` (20), `YarnCliScraper` (20), `PackerCliScraper` (declaration-relative)
- regex `IsMatch` + threshold: `AnsibleCliScraper`, `GradleCliScraper` (4), `JqCliScraper` (20), `LiquibaseCliScraper` (20)

Also worth folding in: the repeatable-value lookahead in `CliScraperBase.HelpDeclaresRepeatableOption` walks continuation lines with `!OptionLinePattern().IsMatch(...)`, a third continuation rule next to the shared one.

### Proposal

Port each scraper to `AccumulateWrappedDescription(lines, ref index, match.Groups["desc"], IsOptionRow)`, passing the scraper's own option regex as the row predicate, and delete the private copies (including the remaining `line.Length - line.TrimStart().Length` indentation hand-rolls). Because each port changes generated output for that tool, do them one tool (or a small group) per PR with a fixture test that pins a wrapped option-looking line, then regenerate the affected package via the workflow after merge.

### Acceptance criteria

- No scraper under `Scrapers/Cli` defines its own multi-line description accumulation or indentation counter.
- Each ported scraper has a fixture test where a wrapped description line beginning with `--flag` stays part of the description.
- `ModularPipelines.OptionsGenerator.Tests` passes.

thomhurst:
[...]

PR #4645, opened 2026-09-05
Share column-aware wrapped-description detection across CLI scrapers

thomhurst:
## Summary

Several CLI scrapers each carried a private "is this line a wrapped description continuation?" rule (Cargo indent-based, Git single-line and dash-rejecting, Cobra hard-coded 20-space threshold) and re-derived `GetIndentation` locally. This PR hoists one shared, column-aware mechanism onto `CliScraperBase` and collapses those variants onto it:

- `GetIndentation` / `GetColumn` expand tabs to eight-column stops (the width the terminal used when the help formatter aligned its columns).
- `IsContinuationLine(line, declarationIndentation, descriptionColumn, looksLikeOptionRow)` is the generic rule: blank lines stop; an option-looking row stops unless its content starts at or after the block's description column (wrapped prose such as a wrapped `--flag=value` mention); otherwise prose continues while deeper than the declaration.
- `AccumulateWrappedDescription` joins the inline description with the prose beneath it and advances the caller's index past every consumed line, so wrapped `--flag` tokens are never re-read as declarations. Scrapers only supply their own "is this an option row" predicate.

Collapsed onto the helper: Cargo, Git, Cobra (and therefore all 19 Cobra-based tools), Hadolint. Flyway, Snyk and `CliArgumentGroupParser` now use the shared indentation primitive.

## Behavioural notes

- Cobra: the 20-space continuation threshold and the `IsSectionHeader` stop rule are gone. Sections are already cut at column-0 headers, and same-depth lines fail the depth rule, so the remaining cases were wrapped prose. Descriptions that previously truncated at a tab-indented continuation (kubectl style) or at a wrapped line beginning with `--flag` now join correctly.
- Cargo: a wrapped line consisting of an option-looking token (`--jobs`) no longer becomes a phantom option.
- Git: descriptions wrapped across several lines are now joined (previously only one wrapped line was read).
- Brew is not touched here: its column-aware rule lives in #4635, which is still open. Once #4635 merges, `BrewCliScraper.IsWrappedDescriptionLine` can be replaced by the shared helper.
- Per the issue's scope note this changes generated output for tools whose wrapped descriptions currently mis-parse; those packages should be regenerated via the workflow after merge.

## Follow-up

[...]

PR #4658, opened 2026-09-06
Port the remaining nine CLI scrapers onto the shared wrapped-description helper

### LINKED ISSUE #4654: Port Az, DotNet, Go, Maven, pip, pnpm, Terraform, WinGet and Yarn scrapers onto AccumulateWrappedDescription
Part of #4646.

These nine scrapers stop accumulating at any dash-led line or below an absolute indent threshold (8-30 spaces). Collapse them onto `CliScraperBase.AccumulateWrappedDescription` with each scraper's option regex as the row predicate so wrapped prose that mentions `--flag` is no longer dropped.

### Acceptance criteria
- None of the nine scrapers defines its own multi-line description accumulation or indentation counter.
- Each has a fixture test where a wrapped description line beginning with `--flag` stays part of the description.
- `ModularPipelines.OptionsGenerator.Tests` passes.
- Regenerate the affected packages via the workflow after merge.

thomhurst:
## Summary

Second child of #4646. Ports the remaining nine scrapers onto `CliScraperBase.AccumulateWrappedDescription` (from #4645): **Az, DotNet, Go, Maven, pip, pnpm, Terraform, WinGet, Yarn**. Their private accumulators (dash-led stops, absolute indent thresholds of 8–30 spaces, and per-tool section-header heuristics) and Go's private `GetIndentationWidth` are deleted; each scraper now supplies only its own option-row predicate. Every ported scraper has a fixture that pins a wrapped `--flag`-led line as prose (Go's existing `Preserves_Hyphen_Led_Description_Continuations` already covers it).

Base: a row whose prose only begins on the next line (go, picocli, argparse) had no description column, so the column-aware rule could not protect its later wrapped lines. `AccumulateWrappedDescription` now infers the column from the first wrapped line. The same four-line hunk is in #4657 (deliberately identical so either PR can land first; git merges it as a no-op), and it is pinned here by two direct `ContinuationLineTests` cases.

## Behavioural notes

- The dropped section-header stops were redundant: every option loop is already section-scoped before accumulation, and no real sub-header sits deeper than its option rows in these nine formats.
- **Az:** knack prints an argument's long summary and preview notice four columns under the row (`Argument '--x' is in preview…`); the old `< 20` threshold dropped those lines, the depth rule now joins them into the description. Pinned in `AzCliScraperTests`.
- Per the parent issue's scope note this changes generated output for these tools wherever wrapped descriptions previously mis-parsed; regenerate the nine packages via the workflow after merge. The base lookahead in `HelpDeclaresRepeatableOption` is tracked in #4655.

## Test plan

- [x] New fixtures for Az, DotNet, Maven, pip (new `PipCliScraperTests`), pnpm (new `PnpmCliScraperTests`), Terraform, WinGet, Yarn; two direct base tests for the inferred column.
- [x] `ModularPipelines.OptionsGenerator.Tests`: 1328 passed, 0 failed (guarded local run).
- [x] `dotnet format whitespace --verify-no-changes` clean for the touched files (the only remaining hit, `IHelpTextCache.cs`, pre-exists on `main`).

Closes #4654

https://claude.ai/code/session_01PkLNTUfwGXjqXZrYrHaDGC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command-line help parsing across supported tools.
  * Wrapped descriptions that resemble option rows are now preserved as prose instead of being incorrectly parsed as separate options.
  * Multi-line descriptions are more reliably combined, including continuation text with varying indentation.

* **Tests**
  * Added coverage for wrapped descriptions and continuation-line parsing across multiple CLI formats.

### DIFF
--- a/tools/ModularPipelines.OptionsGenerator/src/ModularPipelines.OptionsGenerator.Tests/Scrapers/AzCliScraperTests.cs
+++ b/tools/ModularPipelines.OptionsGenerator/src/ModularPipelines.OptionsGenerator.Tests/Scrapers/AzCliScraperTests.cs
@@ -661,6 +661,35 @@ public class AzCliScraperTests
         }
     }
 
+    [Test]
+    public async Task Wrapped_Descriptions_That_Mention_Flags_Stay_Prose()
+    {
+        const string helpText = """
+            Command
+                az vm create : Create an Azure Virtual Machine.
+
+            Arguments
+                --name -n [Required] : Name of the virtual machine. The operation returns immediately when
+                                       --no-wait is also set.
+                --size               : The VM size to be created.
+                    Argument '--size' is in preview and under development.
+            """;
+
+        var command = await new TestAzCliScraper().Parse(["az", "vm", "create"], helpText);
+
+        using (Assert.Multiple())
+        {
+            await Assert.That(command!.Options.Select(option => option.SwitchName))
+                .IsEquivalentTo(["--name", "--size"]);
+            await Assert.That(command.Options.Single(option => option.SwitchName == "--name").Description)
+                .IsEqualTo("Name of the virtual machine. The operation returns immediately when --no-wait is also set.");
+
+            // knack prints preview notices four columns under the row; they now join the description.
+            await Assert.That(command.Options.Single(option => option.SwitchName == "--size").Description)
+                .IsEqualTo("The VM size to be created. Argument '--size' is in preview and under development.");
+        }
+    }
+
     private sealed class TestAzCliScraper()
         : AzCliScraper(
             new ProcessCliCommandExecutor(NullLogger<ProcessCliCommandExecutor>.Instance),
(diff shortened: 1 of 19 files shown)


thomhurst:
Addressed the finding in 932832d626: `IsContinuationLine` now treats an unknown description column as `accept` (`descriptionColumn is null || indentation >= descriptionColumn`), so the first wrapped line can establish the column even when it matches the scraper's option pattern. The null-column `ContinuationLineTests` case is flipped to the corrected behaviour (and still rejects a line no deeper than the declaration), and a new direct `AccumulateWrappedDescription` case pins a first wrapped line starting with `--namespace` being joined while a shallower real option row still ends the description. Full `ModularPipelines.OptionsGenerator.Tests` run: 1329/1329.

@claude please re-review exact head 932832d62685d199bc1a357e0c6ded335a90960f.

PR #4709, opened 2026-09-06
ci: install podman from Homebrew for CLI option generation

### LINKED ISSUE #4690: Generation resolves apt podman 4.9.3 and reports the 5.8.4 command set as removed
## Summary

The scheduled Generate CLI Options run 34006412119 (#4680) failed for podman in `Run Generator for podman`:

```
System.InvalidOperationException: Command coverage validation failed for podman:
  - Removed: podman artifact, podman artifact add, podman artifact extract, podman artifact inspect, podman artifact ls, podman artifact pull, podman artifact push, podman artifact rm, podman machine cp, podman machine reset, podman quadlet
  - Violations: Removed commands require explicit approval: podman artifact, podman artifact add, ... podman machine cp, podman machine reset, podman quadlet
```

This is a real version regression on the runner, not a scraper bug. The job log records:

```
- podman (podman version 4.9.3): 239 commands
  - Baseline comparison: 255 commands at podman version 5.8.4 -> 239 commands at podman version 4.9.3
```

`src/ModularPipelines.Podman/Generated/Podman.Generation.json` on `main` records `"toolVersion": "podman version 5.8.4"`. The run resolved Ubuntu's apt podman 4.9.3, which predates `podman artifact` (5.4) and `podman quadlet` (5.6), so those commands legitimately do not exist there and the coverage guard did its job.

## What to decide

- Make the tool catalog / generation workflow install a podman at least as new as the snapshot (5.8.x) instead of falling back to the runner image's apt package, so scheduled runs cannot shrink the command set by resolving an older tool.
- If a current podman cannot be installed reliably on the runner, decide the supported version explicitly and approve the shrinkage through the existing `--approve-command-coverage-shrinkage` path in one deliberate PR. Do not approve it as a side effect of a scheduled run.

Related: #4680 (tracker), #4667 (phantom removals from help timeouts, a different cause). Filed by the issue/PR loop on 2026-09-06.

thomhurst:
## Summary

The scheduled podman regeneration (#4680, run 34006412119) resolved Ubuntu's apt `podman` 4.9.3 and the coverage guard rejected eleven removed commands (`podman artifact *`, `podman machine cp/reset`, `podman quadlet`). Those commands are real 5.4+/5.6+ additions, so the guard was right: the runner was scraping an older tool than the 5.8.4 snapshot on `main`.

Root cause: the job log for the run that produced the 5.8.4 snapshot (33748284437, image `ubuntu-24.04` 20260823.283.1) shows apt installing 4.9.3 but the scraper pinning `/usr/local/bin/podman`, a newer podman the runner image happened to ship. Image 20260831.293.1 no longer ships it, so `command -v podman` now resolves `/usr/bin/podman` 4.9.3.

## Change

- Install podman from the Homebrew bottle (the Linux formula is the full podman with conmon/crun, not the remote-only client) instead of apt, and prepend `/home/linuxbrew/.linuxbrew/bin` to `GITHUB_PATH` so the "Pin resolved executable" step picks it up.
- Add `podman` to the "Verify affected tool installation" list so the resolved version is printed in the job log.

The next podman regeneration will move the snapshot from 5.8.4 to the current Homebrew release (6.1.1 today); that is additive relative to the baseline and is what the coverage guard expects.

## Validation

A podman-only `workflow_dispatch` of this branch with `create-pr=false` is linked in the comments; it exercises the install step and the coverage comparison end to end.

## Not in this PR

- A generic "resolved tool is older than the snapshot" guard in the generator would catch this class of drift for every tool, but tool version strings are not uniformly parseable; the coverage guard already fails closed on removals, which is what caught this run.
- Buildah and Skopeo still install from apt; their snapshots were produced from apt, so they are consistent today.

Closes #4690

https://claude.ai/code/session_01PkLNTUfwGXjqXZrYrHaDGC

### DIFF
--- a/.github/workflows/generate-cli-options.yml
+++ b/.github/workflows/generate-cli-options.yml
@@ -419,8 +419,13 @@ jobs:
           fi
           cp "$compose_plugin" "$CLI_TOOLS_ROOT/bin/docker-compose"
           chmod +x "$CLI_TOOLS_ROOT/bin/docker-compose"
-          sudo apt-get update
-          sudo apt-get install -y podman
+          # Ubuntu's apt package is podman 4.9, which predates `podman artifact` (5.4) and
+          # `podman quadlet` (5.6), so regenerating from it shrinks the command set. The
+          # runner image no longer ships a newer podman either, so install the current
+          # Homebrew bottle and put it ahead of /usr/bin for the remaining steps.
+          brew="/home/linuxbrew/.linuxbrew/bin/brew"
+          HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_INSTALL_CLEANUP=1 "$brew" install podman
+          echo "/home/linuxbrew/.linuxbrew/bin" >> "$GITHUB_PATH"
           echo "PODMAN_COMPOSE_PROVIDER=$CLI_TOOLS_ROOT/bin/docker-compose" >> "$GITHUB_ENV"
 
       - name: Install Buildah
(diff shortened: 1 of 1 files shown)


thomhurst:
Validation run (podman only, no PR): https://github.com/thomhurst/ModularPipelines/actions/runs/34019651118

Issue #5115, opened 2026-09-14
Land Cargo and shared synopsis source before authoritative regeneration


thomhurst:
Parent: #4692. Existing implementation: #4722 at 75f00f41e0a8bd2f303e4dacba33f7cbb9d7a5de.

The Cargo/shared-synopsis fixes in #4722 cannot land with its current generated output: Linux CI run 34845397917, job 103981055719 fails CS1729 in StateIdentitiesOptionsTests.cs(15,27), because the committed TerraformStateIdentitiesOptions has no required boolean constructor. Cargo, Docker, Syft, Vault, and Yq snapshots also have stale source provenance. Generated files must be regenerated from main after the source changes land.

Deliver the generator/scraper source and generator tests independently:

- Preserve custom Cargo clap headings and resolve positional identity, option-attached values, inherited flag shapes, and requiredness consistently across CLI adapters. Preserve adapter normalization and suppress unsupported conjunctive alternatives instead of flattening them into incorrect OR validation.
- Resolve pnpm alternative shapes before required flags; infer the required Terraform state identities JSON flag from the authoritative synopsis while retaining optional address filters.
- Probe GitHub extension inventory asynchronously before granting optional stack-subtree coverage exemptions. Reject unreliable inventory and partial subtree discovery. Keep publication atomic on probe or coverage failures.
- Preserve main's independent enhancement circuit breaker, process deadlines, AWS concurrency, and Terraform Stacks serialization. Validate the scoped generator solution and relevant tests under the standard guard.

Acceptance: merge a source-only PR with these generator tests and no generated snapshots or integration tests requiring ungenerated APIs. Parent #4692 and #4722 remain open for authoritative latest-main generation of all 47 catalog entries, fresh API/provenance verification, and enabled matching integration tests. Do not weaken, skip, or hand-edit generated output to satisfy the old contracts.

thomhurst:
Opened #5116 with the generator source and tests from #4722, based on current main c107d8cdec. All 934 scraper tests and 77 coverage/orchestration/timeout/enhancement tests pass. The Release generator solution build has zero warnings/errors; scoped default formatting verification, diff checks, and 61 public API baselines pass. Existing severity-info diagnostics and the prior broad compiled-test memory limit are documented.

Parent #4692 and #4722 remain open for authoritative latest-main regeneration and enabled integration contracts. No generated snapshot or integration test requiring an ungenerated API is included in the source PR.

Issue #4692, opened 2026-09-06
Regenerated CargoAddOptions breaks CargoOptionsTests so the cargo update cannot be opened


thomhurst:
## Summary

The scheduled Generate CLI Options run 34006412119 (#4680) failed for cargo in `Build solution to verify changes`, after the generator itself succeeded:

```
test/ModularPipelines.Rust.UnitTests/CargoOptionsTests.cs(15,45): error CS7036: There is no argument given that corresponds to the required parameter 'Dep' of 'CargoAddOptions.CargoAddOptions(...)'
test/ModularPipelines.Rust.UnitTests/CargoOptionsTests.cs(25,63): error CS0117: 'CargoAddOptions' does not contain a definition for 'Path'
test/ModularPipelines.Rust.UnitTests/CargoOptionsTests.cs(35,63): error CS0117: 'CargoAddOptions' does not contain a definition for 'Git'
```

The regenerated `CargoAddOptions` now takes the required `Dep` operand in its constructor and no longer exposes `Path` / `Git` under those names, so the hand-written tests in `ModularPipelines.Rust.UnitTests` no longer compile against current cargo output. Because the build check fails, the workflow never opens the automated cargo PR, so the regeneration cannot land and the tests cannot be fixed on `main` against it.

## Proposal

- Regenerate cargo from current `origin/main` (workflow dispatch, or a local run of the generator with provenance recorded the same way the workflow does) and update `CargoOptionsTests.cs` to the regenerated API in the same PR: construct `CargoAddOptions` with its `Dep` operand and use the current member names for the path and git sources.
- Consider whether hand-written tests for generated option shapes should assert through a stable surface (for example the rendered argument list) rather than constructor signatures, so a legitimate regeneration does not block itself.

Per CLAUDE.md the current tool output is the sole source of truth; the tests follow the generated API, not the reverse.

Related: #4680 (tracker). Filed by the issue/PR loop on 2026-09-06.

thomhurst:
The remaining Terraform integration failure on #4722 is still the stale generated constructor: https://github.com/thomhurst/ModularPipelines/actions/runs/34892639402/job/104140016707 reports CS1729 for TerraformStateIdentitiesOptions(true). The required-JSON generator source is already merged through #5116.

[...]

PR #5119, opened 2026-09-14
fix(generator): preserve Vault switch identity and sync source API

thomhurst:
Vault help uses single-dash switches, but the scraper changed them to double-dash names. Shared synopsis resolution could then discard a positional operand after a presence-only flag and omit the command during traversal. Preserve the documented switch spelling, retain those operands across alternative synopses, and keep value-option placeholders excluded.

Synchronize the generator public API baseline with the source that landed in #5116, including the removed static GetPositionalArguments overloads and the new instance method. The repository sync script recovered the stale entries; an explicit entry also records the linked CommandLinePhase.LateOperand member omitted by its package-path filter.

This follow-up was prepared during review of #5116, which merged before the fixes were pushed. It contains scraper source, tests, and the generator API baseline. Parent #4692 remains open for latest-main generation and integration verification.

Validation: all 1,088 scraper tests pass, including three regressions that failed before the fix. Release generator solution build with PublicApiAnalyzerProject=ModularPipelines.OptionsGenerator passes with zero warnings/errors. Scoped formatting and git diff --check pass. The commit was replayed onto main after #5116 merged; its resulting tree is identical to the validated tree. The existing 600-second/2-GB limits remain unchanged; previously memory-limited broad compiled-generator validation remains deferred to CI.

Refs #5115, #4692

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Vault CLI option parsing to preserve original flag names and prevent duplicate options.
  * Corrected positional argument handling when flags appear before operands.
  * Continued excluding option-value placeholders from positional arguments.
* **Tests**
  * Added coverage for standard and alternative Vault command syntaxes, including help output handling.


thomhurst:
Rebased onto main 8ea298f9b4 after #5118 merged. Resolved the public API header conflict while retaining the shared enum API additions. Remote head 7840fc912fbe822c82fb59e36518e5bed37cd2a2 passes the Release generator solution build with PublicApiAnalyzerProject=ModularPipelines.OptionsGenerator explicitly enabled (zero warnings/errors), all five Vault regressions, and git diff --check. Source formatting remains unchanged. @codex review

thomhurst:
Merged as 34b8b189d15e6cd5054044cc0a5e765c06260b5d after current-head reviews cleared and all 17 checks passed or completed with allowed skipped/neutral conclusions. Paginated branch-run inspection confirms no queued, waiting, or running CI remains, including superseded commits.

PR #5124, opened 2026-09-14
[Automated] Update pip CLI Options

thomhurst:
## Summary
This PR contains automatically generated updates to **pip** CLI options classes.

The generator scraped the latest CLI help output from the installed tool.

## Changes
- Updated options classes to reflect latest CLI documentation
- Added new commands if any were detected
- Updated option types and descriptions

## Assembly-wide public API impact

Affected API families: `Pip`.

- Added APIs: 8
- Removed or changed APIs: 10
- Members with matching names but changed signatures: 6

Breaking changes are present. Consumers may need to update method arguments, option property types or nullability, enum members, and references to removed APIs.

Representative removed or changed members:
- `ModularPipelines.Python.Options.PipUninstallOptions.Deconstruct(out System.Collections.Generic.IEnumerable<string!>! Package) -> void`
- `ModularPipelines.Python.Options.PipUninstallOptions.Package.get -> System.Collections.Generic.IEnumerable<string!>!`
- `ModularPipelines.Python.Options.PipUninstallOptions.Package.init -> void`
- `ModularPipelines.Python.Options.PipUninstallOptions.PipUninstallOptions(System.Collections.Generic.IEnumerable<string!>! Package) -> void`
- `ModularPipelines.Python.Options.PipWheelOptions.Deconstruct(out System.Collections.Generic.IEnumerable<string!>! RequirementSpecifier) -> void`

Representative added members:
- `ModularPipelines.Python.Options.PipUninstallOptions.Package.get -> System.Collections.Generic.IEnumerable<string!>?`
- `ModularPipelines.Python.Options.PipUninstallOptions.Package.set -> void`
- `ModularPipelines.Python.Options.PipUninstallOptions.PipUninstallOptions() -> void`
- `ModularPipelines.Python.Options.PipWheelOptions.PipWheelOptions() -> void`
- `ModularPipelines.Python.Options.PipWheelOptions.RequirementSpecifier.get -> System.Collections.Generic.IEnumerable<string!>?`

## Command coverage
Command coverage report:
- pip (pip 24.0 from /usr/lib/python3/dist-packages/pip (python 3.12)): 14 commands, tree a9b92993f96a5cee7b1131d5da44c679e8bd64ee6a736a1d92dc95e915bd18ac
  - Baseline comparison: 14 commands at pip 24.0 from /usr/lib/python3/dist-packages/pip (python 3.12) -> 14 commands at pip 24.0 from /usr/lib/python3/dist-packages/pip (python 3.12)

## Verification
[...]

PR #5125, opened 2026-09-14
fix(python): support pip requirement-file input alternatives

### LINKED ISSUE #4737: Generator: pip requirement-file alternatives must not require positional operands
Found while auditing generated options PR #4554 at ea5c3056ca59427deef95e7116bc3e887a3bfdd1.

`PipUninstallOptions.Package` and `PipWheelOptions.RequirementSpecifier` become constructor operands with `CliArgument(... Required = true)`. This rejects valid requirement-file-only invocations. For example, `new PipUninstallOptions([]) { Requirement = ["requirements.txt"] }` satisfies the generated Package-or-Requirement validator but `CommandArgumentBuilder.AddArguments` then rejects the empty required positional. Wheel has the same hard-required positional problem.

Upstream documents requirement-file-only forms: [pip uninstall](https://pip.pypa.io/en/stable/cli/pip_uninstall/#usage), [pip wheel](https://pip.pypa.io/en/stable/cli/pip_wheel/#usage).

Fix the shared usage/alternative parsing and requiredness metadata in OptionsGenerator; do not hand-edit generated options or add compatibility shims. #4722 addresses option-attached operands in `UsageSynopsisParser` for #4692 and may provide the underlying fix, but its current tests/regeneration cover cargo rather than these two pip runtime cases. This issue tracks the pip-specific verification and regeneration so it is not lost when the cargo issue closes.

Acceptance criteria:
- Add pip scraper fixtures for uninstall and wheel usage alternatives with `-r <requirements file>`; option operands must not count as positional slots.
- Generated positional requiredness and alternative validation agree.
- Build command lines for requirement-file-only uninstall and wheel without exceptions or dummy operands; assert each emits the intended `-r`/`--requirement` argument.
- Positional-only forms still work; invalid empty input follows the actual CLI contract.
- Regenerate pip from current main after the generator fix. No handwritten generated-output patches.

thomhurst:
Pip uninstall and wheel accept requirement files without dummy positional operands. Wheel also accepts editable-only input, while calls with no input are rejected. Single-use collections retain every input across validation and repeated command rendering.

This PR includes authoritative pip output from #5144 (generation run 35001246869, based on main 1b1f082708), the Python friend-assembly grant, and runtime regressions. Generated files are byte-identical to #5144. The generator snapshot fix landed separately in #5128; no generated options were hand-edited.

Validation: Python Release solution build passes with zero warnings/errors. All 20 Python tests pass, including 12 single-use cases across wheel/uninstall, positional/requirement-file inputs, and empty/one/two elements. Scoped severity-info formatting passes. Tests render commands without executing pip.

Closes #4737.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for pip build input validation and command-line rendering.
  * Verified support for requirement files, positional requirements, editable projects, and single-use inputs.
  * Added validation coverage for missing or empty pip inputs.


thomhurst:
Generated base PR #5124 was closed after review found an additional generator defect: alternative validation consumes single-use collection inputs before rendering. #5126 now blocks parent #4737. These eight tests remain valid for ordinary input collections; this PR stays open and will gain single-use runtime coverage and be rebased onto fresh generated output after the source fix merges. No generated files will be patched in place.

thomhurst:
The single-pass input finding is confirmed and is fixed in source PR #5128 (bff3b56c4850dcb05160ffbd9a21f9c10bfcff9b). It remains outstanding here until that source merges, pip is freshly regenerated, and this PR adopts the replacement output plus single-use command-rendering tests. The current main-based diff still contains the older generated output, so this PR must not merge yet. The domain-collection concern does not reproduce on current main de11378a20: Required_Domain_Collections_Compile_And_Retain_Values, Required_Domain_List_Without_Collection_Metadata_Rejects_Empty_And_AllNull_Input, and Required_Map_Options_Generate_Collection_Guards pass, including the strict generator build and all 1,189 scraper tests. Source #5128 also compiles optional IReadOnlyList<KeyValue> and List<KeyValue> alternatives successfully. These tests exercise generation and compiled behavior rather than assuming the probe resolves the domain type.

thomhurst:
Fixed the single-pass input finding from reviews 5670378300/5670471490 in 633503d8d82e16b6495037fee464930bf2e463b8, verified at the remote PR head. The branch now includes authoritative output from #5144, generated from main after source fix #5128. All 20 Python tests pass, including 12 cases proving that single-use inputs retain every value across validation and two command renders; empty inputs remain rejected. Release build has zero warnings/errors and scoped severity-info formatting passes. Generated files exactly match #5144. The earlier domain-element resolver concern was addressed by the already-merged source; this regeneration successfully traversed that generator and produced the complete 14-command pip output. Additional validation: all 70 AWS scraper tests pass on this source, including Required_Map_Options_Generate_Collection_Guards, which the earlier domain-element finding predicted would throw.

Issue #5133, opened 2026-09-14
AWS rendering fixture omits required app name and blocks regeneration


thomhurst:
The fresh AWS generation run https://github.com/thomhurst/ModularPipelines/actions/runs/34894162179/job/104144911096 completed scraping and built the regenerated AmazonWebServices library. Its solution build then failed before PR creation:

`AwsAmplifyCreateAppOptionsTests.cs(12,44): error CS7036: There is no argument given that corresponds to the required parameter Name of AwsAmplifyCreateAppOptions.AwsAmplifyCreateAppOptions(string)`

The handwritten CreateApp_Joins_Environment_Variables fixture currently constructs an app without its required name. The source fix #5104 correctly makes Name a constructor requirement during regeneration. Update the fixture to supply a complete named app while preserving the environment-variable joining assertion. Use a fixture construction path that works with the current generated class and the forthcoming required constructor, so the test fix can land before regeneration; do not add generated API shims or edit generated output.

Acceptance:
- The rendering fixture supplies an app name and still checks that environment variables render as a comma-joined option value.
- The fixture works with the current class and the generated required constructor.
- A fresh Generate CLI Options run from main can pass AWS solution compilation and create its replacement PR.

Parent #4337; runtime constructor regressions remain in #4367.

thomhurst:
Fixture fix is ready for review in #5134 at f10be4da74e0fa36687ef697d788638ab97d7d32. Fresh AWS generation remains required after it merges; this issue stays open until that acceptance check succeeds.

thomhurst:
The handwritten fixture fix landed in #5134 (d0e82098bc755a017caad9066890404ba4fd6bfe). Fresh AWS regeneration is now running from that main commit: https://github.com/thomhurst/ModularPipelines/actions/runs/34908665241 . PR creation is enabled, auto-merge is disabled, and coverage-shrinkage approval is disabled. This issue stays open until generation and the affected integration checks succeed.

thomhurst:
Run 34908665241 passed the regenerated AWS solution build and created #5140, confirming the fixture compilation fix. #5140 was then closed because main advanced through #5109 during the long generation run: Fast fail job 104215826395 correctly rejected stale generator provenance, so integration tests did not execute. This issue stays open for a fresh generation and integration pass after the pending generator source fixes settle. Parent #4337 and wider exact-source regeneration #4474 retain that acceptance work.

thomhurst:
[...]

Issue #4337, opened 2026-08-30
AWS scraper: preserve required option markers


thomhurst:
## Problem

AwsCliScraper hard-codes every named option as optional, discarding AWS help required markers. Generated services therefore allow invalid empty calls such as EC2 terminate-instances without --instance-ids or create-key-pair without --key-name.

## Acceptance criteria

- Capture required markers from AWS synopsis and option documentation.
- Generate required constructor members and non-null service facade parameters.
- Preserve compatibility only for shipped APIs.
- Add regression tests for EC2 terminate-instances and create-key-pair.
- Regenerate affected AWS options and facades.

Discovered while validating #4311.

thomhurst:
Claimed for implementation. I will preserve AWS required-option markers through parsing/generation, add focused EC2 regressions, and regenerate only affected artifacts through the production generator.

thomhurst:
Both source prerequisites are merged, including required-input validation in #5104. Started fresh AWS generation from main de11378a201b04e3ea363a41499711b572471776: https://github.com/thomhurst/ModularPipelines/actions/runs/34894162179. Auto-merge and command-coverage shrinkage approval are disabled. Existing manual PR #4367 now fails its Windows build because AwsAmplifyCreateAppOptionsTests expects a required constructor that its stale generated output does not contain (job 104140007043, CS1729). The fresh generated output must be reviewed and built before the dependent runtime tests can merge. Current repository guidance makes tool output authoritative; no compatibility shim or generated API baseline will be added.

thomhurst:
AWS regeneration https://github.com/thomhurst/ModularPipelines/actions/runs/34894162179 completed scraping and built the regenerated AmazonWebServices library successfully. The pre-PR solution build then failed in the handwritten environment-variable rendering test at AwsAmplifyCreateAppOptionsTests.cs:12: CS7036, missing required Name for AwsAmplifyCreateAppOptions(string). No replacement PR was created.

The generated required constructor is the expected result of #5104. The existing test fixture on main still constructs an unnamed app. A separate test-fixture fix is needed before fresh regeneration can pass the pre-PR build. The generated output will not be patched by hand.

thomhurst:
[...]

PR #5134, opened 2026-09-14
test(aws): supply app name in environment variable fixture

thomhurst:
The AWS generation workflow now emits the correct required Name constructor, but its pre-PR build fails because the environment-variable rendering test constructs an unnamed app. Supply a named JSON fixture and assert that both the name and comma-joined environment variables render correctly. JSON construction accepts both the current parameterless class and the regenerated required constructor, allowing this handwritten test fix to land before regeneration.

Validation: scoped dotnet format and git diff --check pass. Full AWS build/tests remain deferred to CI after the previously documented local 2 GB build limit; that expensive build was not retried. Fresh AWS generation must run after this fix merges to verify the updated constructor and create the replacement generated PR.

Refs #5133 and #4337. The required-constructor runtime regressions remain in #4367.


thomhurst:
Addressed the optional maintainability suggestion from https://github.com/thomhurst/ModularPipelines/pull/5134#issuecomment-5671800145 in 5432b038562017ee728abbdb89fb6da2961fb91b, confirmed in the remote PR head. A comment now explains why JSON supplies constructor inputs when regeneration makes CLI options required. Scoped formatting and git diff --check pass; no behavior changed in this follow-up. @codex review

Issue #4474, opened 2026-09-01
Regenerate placeholder documentation across CLI integrations


thomhurst:
PR #4468 updates the shared usage-synopsis parser to preserve canonical CLI placeholder notation, but its acceptance criteria and checked-in regeneration are scoped to Chocolatey. Other generated integrations therefore remain stale until regenerated; for example, `pnpm add <name>` now models `The <name> operand.` while `PnpmAddOptions.Generated.cs` still contains `The name operand.`.

A local 20-way regeneration already exceeded the mandated 2 GB agent guard, so complete this as a serialized/per-tool fast-follow after #4468 merges.

Acceptance criteria:
- Identify every tracked integration whose generated operand documentation changes under the placeholder-preserving parser.
- Regenerate those integrations from current tool output; do not hand-edit generated files.
- Commit all affected generated options/services/docs atomically or in clearly linked per-tool PRs.
- Verify deterministic regeneration and each affected tool solution within agent limits; leave exhaustive validation to CI where necessary.
- Confirm no unrelated command-surface drift is approved as part of the documentation refresh.

Follow-up from #4452 / #4468 and review discussion r3902977469.

thomhurst:
Started the placeholder-documentation regeneration from exact `main` head `68c249cde1` on coordination branch `issue-4474-placeholder-docs`.

Authoritative workflow run: https://github.com/thomhurst/ModularPipelines/actions/runs/33507680640

The run covers all 46 non-AWS tools currently included in the generation matrix, with command-coverage shrinkage approval disabled and auto-merge disabled. Changed integrations will open linked per-tool PRs against the coordination branch; unchanged integrations will produce no PR. AWS is temporarily excluded because #4367 already has an exact-head AWS generation run using the shared automated AWS branch. After that run releases the branch, AWS will be generated into this coordination branch before the final PR to `main`.

Local multi-tool generation is intentionally not retried because the issue and repository guard record that it exceeds the mandated 2 GB limit.

thomhurst:
Recovered coordination branch onto current `main` at exact head `d6aa99ab55f3b3aba020f53fcb905d919e89da62`.

- Obsolete Cargo documentation delta was already present on current `main`, so it was dropped during rebase rather than overwriting newer generated API.
- Flux, Pip, and Pnpm generated commits remain.
[...]

PR #5141, opened 2026-09-15
fix(generator): preserve conjunctions in usage alternatives

### LINKED ISSUE #4909: Generator: preserve conjunctions within required usage alternatives
Found while extending the full-traversal regression fixture for #4722.

Given these usage forms and parsed option metadata (`--file` takes a string; `--verbose` is a presence-only flag):

```text
Usage: cargo run [OPTIONS] --file <FILE>
       cargo run [OPTIONS] --verbose <TARGET>
```

The generated required alternative group contains `File` and `Verbose`. `Target` is optional because it is absent from the first form, but selecting `Verbose` alone satisfies that alternative group. The second form requires both `Verbose` and `Target`; a flat File-or-Verbose group loses that conjunction. This was reproduced in the full generator suite while developing ac71d579fc: the traversal fixture returned `[File, Verbose]` instead of retaining the positional requirement in that branch. This synthetic synopsis demonstrates the grammar independently of Cargo's current command catalog.

`GetRequiredAlternativeMembers` treats an operand attached to a required switch as the switch's value before option metadata is available. Once metadata identifies the switch as a flag, required-alternative inference needs to preserve the independent positional and the branch's conjunction. Merely changing the flat group to File-or-Target would still lose the required flag.

Acceptance criteria:
- Add a full-traversal fixture for a value-option-only form versus a form requiring a boolean switch and a positional operand.
- Preserve all required members within each alternative branch; distinguish a switch value from the positional following a presence-only flag using parsed metadata.
- Generated validation accepts File alone and Verbose plus Target, and rejects empty input, Verbose alone, and Target alone.
- Preserve existing one-member alternative behavior and optional flag cases such as `[--verbose] <TARGET>`.
- Regenerate any affected tool output through the generator; do not hand-edit generated options or add compatibility shims.

Implement after #4692/#4722 settles the shared positional selection and requiredness changes.

thomhurst:
Usage forms such as `--file <FILE>` or `--verbose <TARGET>` lost the flag-and-operand conjunction once parsed metadata identified `--verbose` as a presence-only flag. Generated options could consequently omit the required alternative validation entirely.

Preserve multi-member usage branches through synopsis parsing and command resolution, then generate validation for a complete usage form: `File OR (Verbose AND Target)`. A complete form remains sufficient when another form is partly supplied. Optional `[--verbose] <TARGET>` and existing one-member choices keep their behavior. Recursive member resolution also preserves property identities and discards constraints containing unavailable switches.

Full-traversal fixtures compile generated Cargo options and check empty, flag-only, operand-only, complete, false-flag, and overlapping inputs in both synopsis orders. The required-flag cases failed before the fix. Validation passed: strict Release generator API build with zero warnings/errors, 1,251 scraper tests excluding the memory-heavy nested-group suite, and three compiled gcloud negation controls. Scoped formatting and `git diff --check` passed within the default 600-second/2-GB guard. The full nested-group suite remains with CI because previous local runs exceeded that limit.

Closes #4909

Fresh tool output will be generated from main after this and #5128 merge, alongside the pending all-tool regeneration tracked by #4474. No generated options were hand-edited.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CLI option parsing for usage alternatives that combine flags with required operands.
  - Correctly preserves multi-part alternative branches, including nested groups and positional arguments.
  - Generated option validation now accepts complete usage forms while rejecting incomplete combinations.
  - Improved handling when equivalent or overlapping alternative forms are detected.
- **Tests**
  - Added coverage for different argument orderings and combinations of flags, files, and targets.
  - Expanded validation tests for generated option classes.


thomhurst:
@codex review. The initial review attempt reports Failed with no findings or review submitted. Please retry the review of 6666c28cbe63a1f589a83ba3f41984c41be02822 once.

thomhurst:
The initial Claude review run 34924179719 ended without posting a review. Its execution log reports 11 permission denials. Requesting one replacement review for this infrastructure failure; the existing Codex review cleared 6666c28cbe63a1f589a83ba3f41984c41be02822. CI still gates merging.

Issue #5158, opened 2026-09-16
Generator: AWS repeated option metavariables become positional operands


thomhurst:
## Failure
Fresh generated AWS PR #5157 (head 501bca3d1572b37882c8f1547723a9d3a4f54e6f), generated from main d698c0fe, produces incorrect CloudWatch Logs operands and validation despite a successful build.

- AwsLogsTailOptions requires both GroupName and a bogus Value constructor argument. Its new validation then requires optional LogStreamNames whenever Value is supplied.
- AwsLogsStartLiveTailOptions exposes a bogus Value operand from repeated option metavariables and uses it to activate optional stream-filter constraints.
- AWS syntax `group_name <value>` describes one positional argument. `[--log-stream-names <value> [<value>...]]` describes one optional list option, not a positional bundle.

Evidence: https://github.com/thomhurst/ModularPipelines/actions/runs/35036956372 and https://docs.aws.amazon.com/cli/latest/reference/logs/tail.html / https://docs.aws.amazon.com/cli/latest/reference/logs/start-live-tail.html. Installed AWS CLI 2.32.28 help reproduces the same syntax; the generated snapshot used 2.36.46.

## Expected fix
Fix the responsible scraper/parser in tools/ModularPipelines.OptionsGenerator, with full-scrape and generated-runtime regressions. Preserve real positional arguments and optional/required list metadata without inventing Value operands or requiring optional filters. Merge source fixes before regenerating AWS from latest main. Do not patch generated #5157 in place.

Related maintenance: #4367, #5149. #5157 will be closed under repository generated-output policy; link its fresh replacement after source fixes.

thomhurst:
Source fix pushed in #5159 (b50dce2076). Five reproducing scraper failures now pass; 367 related tests and a compiled runtime test covering both CloudWatch tail commands pass. Fresh generation will follow source merges; keeping this issue open until output is verified.

thomhurst:
Source fix #5159 merged as 97b659218a45a36609e360e50f5ce5178df71b45 with all 17 checks passing. Gcloud source #5156 is also merged. Fresh AWS/gcloud generation uses main 67555c6cec5db9cc0f260ebdf501793ac81ae3f7: https://github.com/thomhurst/ModularPipelines/actions/runs/35047414356. Keeping this issue open until the replacement AWS output is reviewed and validated.

thomhurst:
[...]

PR #5159, opened 2026-09-16
fix(generator): preserve AWS metavariable arity

thomhurst:
## Summary
AWS synopsis notation such as `group_name <value>` and `--names <value> [<value>...]` created extra positional operands. In #5157 those operands also made optional CloudWatch Logs stream filters effectively required.

Normalize these AWS metavariables before shared usage parsing. Explicit repeated-value syntax supplies list arity even when the option's element type is documented as `string`; scalar description prose keeps its existing behavior. The generated CloudWatch tail constructors now require only their actual inputs and accept omitted optional filters.

Related: #5158, closed generated PR #5157, runtime PR #4367. Keep #5158 open until fresh generation succeeds. Merge this source fix and #5156 before regenerating AWS/gcloud from latest main.

## Validation
- Reproduced five failures in six new scraper cases before the fix.
- 367 parser/AWS/gcloud/kubectl tests pass.
- Compiled runtime regression passes for both CloudWatch tail commands, including omitted, empty, and populated stream filters.
- Scoped generator solution formatting passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AWS CLI option parsing for repeated metavariables, ensuring they generate multi-value options instead of positional arguments.
  * Correctly handles named positional metavariables and preserves required constructor validation for AWS log commands.
  * Improved support for optional filters and prefix options in generated AWS CLI options.

* **Tests**
  * Added coverage for AWS CLI synopsis parsing and generated option validation.


thomhurst:
Review dispositions for b50dce207622dc498faa1f207fa09cc17f346362:

- Explicit repeated metavariables describe multiple CLI values; they intentionally take precedence over an element-type label. Scalar/key-value prose still cannot create a list. Fresh AWS output will be checked against the source CLI before merge.
- Normalization and option classification already share AwsOptionMetavariablePattern. They consume it separately because normalization rewrites synopsis text while classification extracts arity from the original help; no duplicated matching rule needs another abstraction.
- CodeRabbit's docstring percentage includes private regex helpers and tests. The new protected override inherits its contract, and comments explain the non-obvious AWS notation. Keeping documentation focused on those contracts rather than adding repetitive test/helper XML summaries.

Both reviews report no actionable findings. Waiting for remaining CI before source merge and regeneration.

Issue #5146, opened 2026-09-15
Generator: gcloud 585 skips 54 commands with operand or repeatability parse failures


thomhurst:
The same latest-main generation run as #5145 also logs parse failures for 54 non-dispatch commands before the fatal colon-group error. These are independent failures in the shared synopsis/gcloud argument parsers and repeatability checks; fixing only #5145 does not establish healthy generated coverage.

Evidence: https://github.com/thomhurst/ModularPipelines/actions/runs/35001246869/job/104490461965, Google Cloud CLI 585.0.0. The attached list excludes the GROUP/COMMAND dispatch-placeholder warnings.

Examples include hyphenated operands (KEY-ID, OUTPUT-FILE), compound SSH/SCP operands, forwarded arguments after --, mismatched declaration/synopsis names, and --add-endpoint being classified as scalar despite repeated-switch documentation.

Expected: parse the current documented operands and collection contracts through generic scraper/parser logic, with captured-help regressions. Do not silently omit these commands, weaken coverage checks, add handwritten generated APIs, or preserve obsolete API shapes. Validate fresh latest-main generation once the source fixes land.

Related runtime PRs: #4863, #4869, #4956. Parent issues: #4849, #4797, #4850.

## Failure evidence

```
gcloud ai custom-jobs local-run declares positional operands in usage 'gcloud ai custom-jobs local-run --executor-image-uri=IMAGE_URI [--extra-dirs=[EXTRA_DIR,...]] [--extra-packages=[PACKAGE,...]] [--gpu] [--local-package-path=LOCAL_PATH] [--output-image-uri=OUTPUT_IMAGE] [--requirements=[REQUIREMENTS,...]] [--service-account-key-file=ACCOUNT_KEY_FILE] [--python-module=PYTHON_MODULE | --script=SCRIPT] [GCLOUD_WIDE_FLAG ...] [-- ARGS ...]', but no CliPositionalArgument values were generated.
[...]

PR #5161, opened 2026-09-16
fix(generator): normalize gcloud synopsis aliases and defaults

thomhurst:
## Problem and change

Gcloud SDK 585 generation on main failed in [run 35047414356](https://github.com/thomhurst/ModularPipelines/actions/runs/35047414356): 24 DNS/SQL commands were skipped because `--instance=INSTANCE, -i INSTANCE` produced a phantom operand, then `builds triggers create manual` aborted generation because a displayed default was interpreted as nested argument syntax.

Normalize comma-separated short aliases with their long option's documented value, and strip displayed default annotations from normalized synopsis options before parsing groups. Original help remains available for option metadata. Preserve real operands after aliases and quoted default text containing syntax characters.

Add 42 captured SDK 585 help fixtures covering every skipped command and all stable build-trigger create/update variants. No generated options are edited.

## Validation

- Before: 25 failures among 97 captured command regressions (24 alias failures plus the fatal manual-trigger group).
- After: 430 parser, gcloud, AWS, and kubectl tests passed, including all 97 captured commands and seven focused alias/default cases and the manual-trigger constraint structure.
- Scoped generator formatting passed with the pinned SDK through the 2 GB agent guard.

Related to #5146, #4863, #4869, and #4956. Keep #5146 open until fresh full generation and coverage pass.

Regenerate gcloud and AWS from latest main after the separate AWS enum defect in #5164 is fixed.


thomhurst:
Addressed both non-blocking observations in review comment 5691564841 with 66033be8456699f01e377d4e0aa84be198f913c6, confirmed in the remote PR head: short aliases now accept both space and equals separators, and the captured manual-trigger regression asserts the required trigger/config choice, nested build-config exclusivity, Dockerfile requirement, and absence of positional operands. All 430 related parser tests pass; scoped formatting passes. The CodeRabbit docstring-coverage warning counts named regression tests and private helpers; these follow existing repository conventions rather than adding repetitive XML comments to test methods.

Issue #5164, opened 2026-09-16
Generator: reject nested AWS enum prose and standalone bullet markers


thomhurst:
Fresh AWS CLI 2.36.46 output in #5163 (head d28ad414dcafe2aa143159dd8ccdba82ee3adcc1) has two correctness defects identified in [review comment 5691970692](https://github.com/thomhurst/ModularPipelines/pull/5163#issuecomment-5691970692). Generation and compilation pass, so these require semantic scraper regressions.

1. `aws sns set-topic-attributes --attribute-name`: `AwsSnsSetTopicAttributesAttributeName` contains prose words (`across`, `deduplication`, `entire`, `is`, `message`, `o`, etc.). The extractor mistakes an embedded sentence describing the values of the nested `FifoThroughputScope` attribute for the value set of `--attribute-name`. The required API cannot express valid names such as `Policy`, `DisplayName`, and `ContentBasedDeduplication`.
2. `aws acm create-acme-endpoint --authorization-behavior`: a one-item `Possible values:` list produces both `EnumValue("o")` and `EnumValue("PRE_APPROVED")`. A lone AWS man-page bullet marker must not become a value. Similar bogus `o` values appear in many generated enums.

Fix the responsible scraper/enum extraction logic in `tools/ModularPipelines.OptionsGenerator/`, preferably generically. Preserve raw structure where needed so nested explanatory prose cannot define an option-level enum. For SNS, retain a string option unless an authoritative complete option-level closed value set is available; do not invent a handwritten enum or API shim. Strip structural bullet markers while retaining the actual values of one-item and multi-item lists.

Acceptance:
- Reproduce both defects with captured current AWS help and focused tests.
- Test embedded/nested prose, genuine option-level enum lists, and single/multiple bullet items.
- Validate generated metadata/rendering for the affected options and run relevant broader parser tests.
- Merge the source fix, then regenerate fresh AWS output from latest main; no repairs to the autogenerated PR in place.
- Confirm healthy replacement coverage/builds/reviews and link it to closed #5163.

Related: #4367, #5149, #5158. The metavariable/ACL fixes in #5150/#5159 are present in this snapshot; these are separate remaining enum defects.

thomhurst:
[...]

PR #5165, opened 2026-09-16
fix(generator): respect AWS enum description boundaries

thomhurst:
## Problem and fix

AWS enum extraction accepted nested attribute prose and treated the single bullet in `Possible values: o PRE_APPROVED` as a second allowed value. In the generated AWS PR #5163 and runtime PR #4367, this made valid SNS attribute names impossible to pass and exposed invalid `O` enum members.

Keep the original description indentation until enum detection. Accept enum headings at the option description level and at the start of a sentence, excluding nested attribute text. Recognize both `o` and `*` bullet lists, including one-item lists, and reject prose within their entries before applying the existing shared enum member limits. A single allowed value therefore retains the existing string fallback.

Regression coverage includes the SNS documentation excerpt, ACM authorization/contact values, the VPN `ipsec.1` value, nested headings, inline and separate option-level headings, and generated constructor/enum output. Generated options are not edited in this PR.

## Validation

- Before the fix: seven regression cases failed in the 101-case AWS scraper run.
- After the fix: all 322 AWS, gcloud, synopsis, description enum, and enum factory tests pass.
- Scoped formatting of the generator solution completed at severity `info`; existing diagnostics without code fixes remain.
- Release build of the generator solution passed with zero warnings and zero errors.

## Regeneration

Fixes the generator portion of #5164. After this source PR merges, regenerate AWS and gcloud together from the latest `main`, review the replacement for closed #5163, and update #4367 with the authoritative output. Keep #5164 open until regeneration is verified.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved AWS CLI option parsing to avoid incorrectly restricting values based on nested attribute documentation.
  - Correctly recognizes option-level “Possible values” sections, including inline and paragraph-separated headings.
  - Preserves complete top-level enum choices while rejecting structural or descriptive bullets as valid choices.
  - Ensures single-value fallback handling remains accurate.
  - Keeps SNS attribute names as required strings and excludes structural bullets from ACM option enums.


thomhurst:
The final head is 1b82365ef330b98f20d9e51832f817e42b71ee57. It also rejects prose inside bullet entries; all 322 related tests pass. @codex review

thomhurst:
Thanks for the review in comment 5692160135. The indentation comparison already rejects any heading deeper than the option description, including more than two indentation tiers. An indentation stack would only be needed if future help formats require interpreting those nested values; this change intentionally keeps the existing string fallback for them. No additional parser state is needed for the current option-level enum contract.

thomhurst:
Disposition of the two pre-merge warnings in comment 5692102410: #5163 is the closed generated PR, not the source issue. This PR implements source issue #5164. Repository guidance explicitly requires merging the source fix separately before regenerating a fresh AWS PR, so generated output belongs in the replacement PR after this merges. The new methods are named regression tests and a private helper; they follow the surrounding test/helper documentation convention, and the helper includes the non-obvious indentation rationale. @coderabbitai review the final head 1b82365ef330b98f20d9e51832f817e42b71ee57, which additionally rejects prose inside bullet entries.

